PR #30, opened 2015-03-09
V2015

nmerket:
@gmdickey @lirainer 

Here's the HPXML translator all updated and ready to go for HEScore v2015. I've done some local testing. It's ready to be dropped into place in the sandbox to see how it works.


nmerket:
@lirainer, have you had a chance to drop this into the v2015 API and test it?

lirainer:
@nmerket. I was holding off on that until after we released the 2015 version since I needed to keep things stable. But now that we are delayed again I think we have time to do it early next week. I will let you know.

PR #69, opened 2018-02-23
Allowing a heating- or cooling-only heat pump

nmerket:
This permits the case where a heatpump is specified but only provides cooling or heating. To specify that a heat pump only provides heating `HeatPump/FractionCoolLoadServed` must be present and equal to `0`. Likewise `HeatPump/FractionHeatLoadServed` must be present and equal to `0` if the heat pump only provides cooling.

This will pass through cases where a heat pump provides heating and a separate air conditioner provides cooling to HEScore. However, it will fail on the HEScore side because it cannot handle that configuration.


bookwyrm:
@nmerket I have updated the AjO code and tested this against a couple different HPXML files and the handling of HeatPump for Heating vs Cooling seems to be working.

👍

Issue #70, opened 2018-04-02
Allow actual 5% difference in heating/cooling fraction load served


bookwyrm:
When processing an HPXML file with `FractionHeatLoadServed` value of 0.95 and `FractionCoolLoadServed` value of 1.0 where the two systems share the same duct system, there is a translation error stating that the two values are not within 5% of each other.

This is likely due to an approximation of floating point values.

I suggest changing the boundary check from `0.05` to `0.051` at https://github.com/NREL/hescore-hpxml/blob/master/hescorehpxml/__init__.py#L1655 in order to allow an actual check of 5% threshold.

nmerket:
Sounds reasonable to me.

PR #71, opened 2018-04-05
Allowing hvac weights to be equal to 5% different

nmerket:
In response to #70.


nmerket:
@bookwyrm, this should solve your problem.

PR #83, opened 2019-03-19
Add solar screens

yzhou601:
Added solar screens to window and skylight


yzhou601:
Sorry for the bad message, the first one is actually the test update. The added test functions passed all cases when ran separately, while some previous tests seemed to be broken when ran the whole test script.

hescorehpxml/__init__.py
```diff
@@ -1101,7 +1101,16 @@ def _get_skylights(self, b):
...
+        skylight_sunscreen_areas = {}
+        for skylight in skylights:
+            solar_screen = bool(xpath(skylight, 'h:Treatments/text()') == 'solar screen'
+                                or xpath(skylight, 'h:ExteriorShading/text()') == 'solar screens')
+            area = convert_to_type(float, xpath(skylight, 'h:Area/text()'))
+            try:
+                skylight_sunscreen_areas[str(solar_screen)] += area
+            except KeyError:
+                skylight_sunscreen_areas[str(solar_screen)] = area
+        zone_skylight['solar_screen'] = json.loads(max(skylight_sunscreen_areas.items(), key=lambda x: x[1])[0].lower())
```
nmerket: You could skip the whole `json.loads` thing by not converting the the dict keys above to strings (i.e. `skylight_sunscreen_areas[solar_screen]`)

```suggestion
        zone_skylight['solar_screen'] = max(skylight_sunscreen_areas.items(), key=lambda x: x[1])[0]
```

hescorehpxml/__init__.py
```diff
@@ -1525,7 +1536,13 @@ def windows_are_on_shared_walls():
...
                         window_code_areas[window['window_code']] = window['area']
                 zone_window['window_code'] = max(window_code_areas.items(), key=lambda x: x[1])[0]
-
+            window_sunscreen_areas = {}
+            for window in windows:
+                try:
+                    window_sunscreen_areas[str(window['sun_screen'])] += window['area']
+                except KeyError:
+                    window_sunscreen_areas[str(window['sun_screen'])] = window['area']
+            zone_window['solar_screen'] = json.loads(max(window_sunscreen_areas.items(), key=lambda x: x[1])[0].lower())
```
nmerket: Same comment here about not converting to strings and not dealing with the json conversion.

tests/tests.py
```diff
@@ -1399,5 +1399,44 @@ def test_roof_absorptance(self):
...
+        self.assertTrue(d['building']['zone']['zone_roof'][0]['zone_skylight']['solar_screen'])
+
+    def test_skylight_solar_screens_ExteriorShading(self):
+        tr = self._load_xmlfile('house4')
+        glasstype = self.xpath('//h:Skylight[h:SystemIdentifier/@id="skylights"]/h:GlassType')
+        el2 = etree.Element(tr.addns('h:ExteriorShading'))
+        el2.text = 'solar screens'
+        glasstype.addnext(el2)
+        d = tr.hpxml_to_hescore_dict()
+        self.assertTrue(d['building']['zone']['zone_roof'][0]['zone_skylight']['solar_screen'])
```
nmerket: I like these tests!

hescorehpxml/__init__.py
yzhou601: Great suggestion, I tried that at the beginning which led to some bugs I thought it might be allowed to use bool as key, so I made it this way. But I retried today it worked. Thanks!

nmerket:
As I mentioned above, the next step is to update the documentation. You can find the latest documentation [here](https://hescore-hpxml.readthedocs.io/en/latest/). We'll need to add a description of this change to the appropriate file in the repository that will generate this documentation. Probably [this file](https://github.com/NREL/hescore-hpxml/blob/solarscreens/docs/source/translation/zone_window.rst). You can find all the documentation files under the `docs` folder in your local clone of the repo. The documents are written in [reStructuredText](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html). That link is a primer on how to write reStructuredText which could be helpful.

nmerket:
To get a live preview of the docs as you edit them:

```
cd C:\path\to\hescore-hpxml\docs
make.bat livehtml
```

nmerket:
@yzhou601 I changed the solar screen docs so it would have its own section in the documentation. This looks great! Merging.

PR #84, opened 2019-03-25
Tankless

### LINKED ISSUE #79: Add tankless water heater
New enumeration for domestic_hot_water/type: tankless

In domestic_hot_water type can be allow tankless if category is unit

 In domestic_hot_water efficiency_method cannot be shippment_weighted if type is tankless.

### LINKED ISSUE #80: Add unified energy factor
New enumeration for domestic_hot_water/efficiency_method: uef

yzhou601:
Added tankless enumeration to output Json schema. If the water heater type is "instantaneous water heater" in HPXML, the domestic_hot_water/type would be "tankless", category is "unit". The efficiency method could only be "user" for this type. If there's no energy factor information, it would cause a translation error.
Two tests are added to check the output type and translation error.

Fixes #79 
Fixes #80


hescorehpxml/__init__.py
```diff
@@ -1890,20 +1890,35 @@ def _get_systems_dhw(self, b):
...
         if not sys_dhw['category'] == 'combined':
             energyfactor = xpath(primarydhw, 'h:EnergyFactor/text()')
-            if energyfactor is not None:
+            unified_energy_factor = xpath(primarydhw, 'h:UniformEnergyFactor/text()')
+            if unified_energy_factor is not None:
+                sys_dhw['efficiency_method'] = 'uef'
+                # Is the uniform energy factor also named "energy_factor"? Does it occur the situation both energy
+                # factors are given? This part automatically gives priority to uef since it's a new metric to replace \
+                # the old one.
+                sys_dhw['energy_factor'] = round(float(unified_energy_factor), 2)
```
nmerket: This is a good question.

@lirainer, if we submit a home with `efficiency_method=uef`, do we then put the UEF in the `energy_factor` field in the API?

hescorehpxml/__init__.py
```diff
@@ -1890,20 +1890,35 @@ def _get_systems_dhw(self, b):
...
             else:
-                dhwyear = int(xpath(primarydhw, '(h:YearInstalled|h:ModelYear)[1]/text()'))
-                if dhwyear < 1972:
-                    dhwyear = 1972
-                sys_dhw['efficiency_method'] = 'shipment_weighted'
-                sys_dhw['year'] = dhwyear
+                # Tankless type must use energy factor method
+                if sys_dhw['type'] == 'tankless':
+                    raise TranslationError(
+                        'Tankless water heater efficiency cannot be estimated by shipment weighted method.')
```
nmerket: Nice job putting a check in for this with a good error message.

hescorehpxml/__init__.py
yzhou601: I can't agree more about the branch suggestion, it used to be in a mess when I wanted to commit and found I was in "solarscreens" all the time! BTW, improper statement in previous comment, the "uef" method is also enabled for tankless water systems.

hescorehpxml/__init__.py
yzhou601: Thanks :)

hescorehpxml/__init__.py
lirainer: @nmerket Yes, it goes in `energy_factor`. Not great, but this allowed us to add UEF as a non-breaking change.

nmerket: @lirainer Not a problem. Just wanted to verify that we're doing it right.

nmerket:
To update the docs:

- Update [this table](https://15-24196988-gh.circle-artifacts.com/0/docs/translation/domestic_hot_water.html) to reflect the new mapping and remove the footnote as it is no longer applicable.
- Change the explanation of water heater efficiency to include UEF.

PR #90, opened 2019-05-03
Allow more hvac combination

### LINKED ISSUE #81: Allow more heating and cooling system combinations
image

The code to do this is [here](https://github.com/NREL/hescore-hpxml/blob/master/hescorehpxml/__init__.py#L1548). We need to

- [ ] See if the translator already permits the combinations above. i.e. write unit tests for them.
- [ ] For ones where it doesn't, update the `get_hvac` method to allow that combination.

yzhou601:
fix #81

### DIFF
--- a/hescorehpxml/__init__.py
+++ b/hescorehpxml/__init__.py
@@ -1554,16 +1554,23 @@ class HPXMLtoHEScoreTranslator(object):
                 return_dict[system_id] = el
             return return_dict
 
+        def remove_hp_by_zero_value(test_variable):
+            if test_variable is not None:
+                test_variable_value = Decimal(test_variable)
+                if test_variable_value == Decimal(0):
+                    return True
+
         # Get all heating systems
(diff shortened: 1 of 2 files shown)


yzhou601:
The json data output looked good after checking three data points: heating fraction/ heating capacity/ heating capacity in 17F

nmerket:
If there are red areas that didn't throw errors that should have, can you add some code to the translation to ensure it does? Also, it looks like there are some other errors happening when I look at the CI results.

yzhou601:
@nmerket Oh yes, that's a red area gave the error message, but it's not descriptive for the problem, I could add a check for that too, and make the right error message.

nmerket:
@yzhou601 Nice work on this! Very thorough testing and it seems like you caught some edge cases that I hadn't before.

yzhou601:
@nmerket No problem, tell me if there's anything else I could help with.

PR #98, opened 2019-07-24
Add cond_attic translation

### LINKED ISSUE #95: Workaround for conditioned attics in HPXML 2.3
There is [no way to translate a conditioned attic](https://hescore-hpxml.readthedocs.io/en/latest/translation/zone_roof.html#id1) from HPXML 2.3. It [will be possible](https://hpxml.nrel.gov/datadictionary/3/Building/BuildingDetails/Enclosure/Attics/Attic/AtticType/Attic/Conditioned) when we move to HPXML 3. We should include an extension to allow people to get that in v2.3. 

Let's do this as follows: To get a conditioned attic in HEScore, you set `Attic/AtticType` to "other" and then `Attic/extension/Conditioned` to "true". If it meets those two criteria, then translate as `cond_attic` in HEScore. If not that act as it normally does (i.e. "other" will cause the translation to fail.).

To Do:

- [ ] Update translation
- [ ] Write test of new capability
- [ ] Update docs to reflect change in behavior

yzhou601:
Allow HEScore conditioned attic to be translated from HPXML 2.3.
Fix #95

### DIFF
--- a/hescorehpxml/__init__.py
+++ b/hescorehpxml/__init__.py
@@ -923,8 +923,8 @@ class HPXMLtoHEScoreTranslator(object):
             hpxml_attic_type = xpath(attic, 'h:AtticType/text()')
             atticd['rooftype'] = rooftypemap[hpxml_attic_type]
             if atticd['rooftype'] is None:
-                attc_is_cond = xpath(attic, 'boolean(h:extension/h:Conditioned)')
-                if attc_is_cond:
+                attc_is_cond = xpath(attic, 'h:extension/h:Conditioned/text()')
+                if attc_is_cond == 'true':
                     atticd['rooftype'] = 'cond_attic'
                 else:
                     raise TranslationError(
--- a/hescorehpxml/__init__.py
+++ b/hescorehpxml/__init__.py
@@ -923,9 +923,13 @@ class HPXMLtoHEScoreTranslator(object):
             hpxml_attic_type = xpath(attic, 'h:AtticType/text()')
             atticd['rooftype'] = rooftypemap[hpxml_attic_type]
             if atticd['rooftype'] is None:
-                raise TranslationError(
-                    'Attic {}: Cannot translate HPXML AtticType {} to HEScore rooftype.'.format(atticid,
-                                                                                                hpxml_attic_type))
+                attc_is_cond = xpath(attic, 'boolean(h:extension/h:Conditioned)')
+                if attc_is_cond:
+                    atticd['rooftype'] = 'cond_attic'
+                else:
+                    raise TranslationError(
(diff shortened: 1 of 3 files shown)


tests/tests.py
```diff
@@ -1970,6 +1970,28 @@ def test_uef_with_tankless(self):
...
+        is_attic_cond.text = 'true'
+        d = tr.hpxml_to_hescore_dict()
+        roof_type = d['building']['zone']['zone_roof'][0]['roof_type']
+        self.assertEqual(roof_type, 'cond_attic')
+        is_attic_cond.text = 'false'
+        self.assertRaisesRegexp(
+            TranslationError,
+            r'Attic \w+: Cannot translate HPXML AtticType other to HEScore rooftype.',
+            tr.hpxml_to_hescore_dict
+        )
```
nmerket: I added this part to test for the case when:

```xml
<Attic>
    <AtticType>other</AtticType>
    <extension>
        <Conditioned>false</Conditioned>
    </extension>
</Attic>
```

which should raise an error.

hescorehpxml/__init__.py
```diff
@@ -923,9 +923,13 @@ def _get_building_zone_roof(self, b, footprint_area):
             hpxml_attic_type = xpath(attic, 'h:AtticType/text()')
             atticd['rooftype'] = rooftypemap[hpxml_attic_type]
             if atticd['rooftype'] is None:
-                raise TranslationError(
-                    'Attic {}: Cannot translate HPXML AtticType {} to HEScore rooftype.'.format(atticid,
-                                                                                                hpxml_attic_type))
+                attc_is_cond = xpath(attic, 'boolean(h:extension/h:Conditioned)')
```
nmerket: It could be fixed by changing this line to check the contents on the element not just whether it exists, which is what it is doing now.

yzhou601: Thanks, I made this change, it should pass CI and the added test now.

nmerket:
Thanks for the quick fix. One more thing, could you add to the test to check if the user has done the following

```xml
<Attic>
    <AtticType>vented attic</AtticType><!-- or whatever -->
    <extension>
        <Conditioned>true</Conditioned>
    </extension>
</AtticType>
```

that it will appropriately ignore the `extension/Conditioned` element?

yzhou601:
> Thanks for the quick fix. One more thing, could you add to the test to check if the user has done the following
> 
> ```
> <Attic>
>     <AtticType>vented attic</AtticType>
>     <extension>
>         <Conditioned>true</Conditioned>
>     </extension>
> </AtticType>
> ```
> 
> that it will appropriately ignore the `extension/Conditioned` element?

It is ignored when attic type is specified, but would we want to throw an error/warning when they're double specified ?

nmerket:
I think ignoring it is okay, as long as we have documented that.

PR #100, opened 2019-08-01
aluminum frame window code mappings

### LINKED ISSUE #96: Window code mapping changes
- Aluminum, double pane, low-e is currently mapping to `dcaa`.
 - Aluminum, thermal break is currently mapping to `dcab`

@lirainer What should these be mapping to?

nmerket:
Fixes #96 by changing the window code mappings for aluminum frame windows.


nmerket:
I should make sure the docs are correct with this change.

Issue #75, opened 2019-03-01
Map Home Performance with Energy Star fields from HPXML


nmerket:
We need to locate and pass a few fields from Home Performance with Energy Star (HPwES) from fields in HPXML to the matching [API inputs for HEScore](http://hes-documentation.labworks.org/home/api-definitions/api-methods/submit_hpwes). These should probably be added to another top level key in the output json since it is separate from `submit_address` and `submit_inputs`

Steps:

- [ ] Make mapping table between [HPXML](https://hpxml.nrel.gov/datadictionary) and [HPwES API inputs](http://hes-documentation.labworks.org/home/api-definitions/api-methods/submit_hpwes)
- [ ] Create new method in [HPXMLtoHEScoreTranslator Class](https://github.com/NREL/hescore-hpxml/blob/master/hescorehpxml/__init__.py#L85) to translate mapping
- [ ] Testing
   - [ ] Update example files
   - [ ] Write tests for HPwES fields
- [ ] Documentation

PR #101, opened 2019-08-19
HPwithES fields mapping from HPXML into HEScore

yzhou601:
#75

### DIFF
--- a/docs/source/translation/hpwes.rst
+++ b/docs/source/translation/hpwes.rst
@@ -1,0 +1,87 @@
+Home Performance with Energy Star
+#################################
+
+Inputs for the Home Energy Score `submit_hpwes`_ API call can now be retrieved
+from an HPXML file.
+
+Project
+*******
+
+To get the Home Performance with Energy Star (HPwES) data
+from an HPXML file a ``Project`` node needs to be included with at least the
+following elements:
+
+.. _submit_hpwes: https://hes-documentation.labworks.org/home/api-definitions/api-methods/submit_hpwes
+
+.. code:: xml
+
+    <Project>
+        <ProjectDetails>
+            <ProjectSystemIdentifiers id="projectid"/>
+            <ProgramCertificate>Home Performance with Energy Star</ProgramCertificate>
+            <StartDate>2018-08-20</StartDate>
+            <CompleteDateActual>2018-12-14</CompleteDateActual>
+            <extension>
+                <isIncomeEligible>true</isIncomeEligible>
+            </extension>
+        </projectDetails>
+    </Project>
+
+If more than one ``Project`` element exists, the first one will be used. The
+user can override this by passing the ``--projectid`` argument to the translator
+command line. To translate the HPwES fields, the ``ProgramCertificate`` must be
+present and equal to ``Home Performance with Energy Star``.
+
+The project fields are mapped as follows:
+
++---------------------------------------+----------------------------------------------+
+|       HPXML ``ProjectDetails``        |          `submit_hpwes`_ API value           |
++=======================================+==============================================+
+| ``StartDate``                         | ``improvement_installation_start_date``      |
++---------------------------------------+----------------------------------------------+
+| ``CompleteDateActual``                | ``improvement_installation_completion_date`` |
++---------------------------------------+----------------------------------------------+
+| ``extension/isIncomeEligible`` [#f1]_ | ``is_income_eligible_program``               |
+|                                       |                                              |
++---------------------------------------+----------------------------------------------+
+
+Contractor
+**********
+
+A ``Contractor`` element is also required with at minimum the following
+elements:
+
+.. code:: xml
+
+    <Contractor>
+        <ContractorDetails>
+            <SystemIdentifier id="contractor1"/>
+            <BusinessInfo>
+                <SystemIdentifier id="contractor1businessinfo"/>
+                <BusinessName>My HPwES Contractor Business</BusinessName>
+                <extension>
+                    <ZipCode>12345</ZipCode>
+                </extension>
+            </BusinessInfo>
+        </ContractorDetails>
+    </Contractor>
+
+If there are more than one ``Contractor`` elements, the contractor with the id
+passed in the ``--contractorid`` command line argument is used. If no contracter
+id is specified by the user, the contractor listed in the
+``Building/ContractorID`` will be used. If that element isn't available, the
+first ``Contractor`` element will be used.
+
+The contractor fields are mapped as follows:
+
++------------------------------------------------------+------------------------------+
+|                 HPXML ``Contractor``                 |  `submit_hpwes`_ API value   |
++======================================================+==============================+
+| ``ContractorDetails/BusinessInfo/BusinessName``      | ``contractor_business_name`` |
++------------------------------------------------------+------------------------------+
+| ``ContractorDetails/BusinessInfo/extension/ZipCode`` | ``contractor_zip_code``      |
++------------------------------------------------------+------------------------------+
+
+.. rubric:: Footnotes
+
+.. [#f1] ``true`` if element value is "true", ``false`` if element value is anything else or missing.
--- a/hescorehpxml/__init__.py
+++ b/hescorehpxml/__init__.py
@@ -670,7 +670,12 @@ class HPXMLtoHEScoreTranslator(object):
         # Create return dict
         hescore_inputs = OrderedDict()
         hescore_inputs['building_address'] = self._get_building_address(b)
-        hescore_inputs['hpwes'] = self._get_hpwes(b, p, c)
+
+        if p is not None:
+            program_c = xpath(p, 'h:ProjectDetails/h:ProgramCertificate/text()')
+            if program_c is not None and program_c == 'Home Performance with Energy Star':
+                hescore_inputs['hpwes'] = self._get_hpwes(b, p, c)
+
         bldg = OrderedDict()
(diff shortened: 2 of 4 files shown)


hescorehpxml/__init__.py
```diff
@@ -626,12 +626,16 @@ def hpxml_to_hescore_json(self, outfile, *args, **kwargs):
...
 
-    def hpxml_to_hescore_dict(self, hpxml_bldg_id=None, nrel_assumptions=False):
+    def hpxml_to_hescore_dict(self, hpxml_bldg_id=None, hpxml_project_id = None, hpxml_contractor_id = None, nrel_assumptions=False):
         '''
         Convert a HPXML building file to a python dict with the same structure as the HEScore API
 
         hpxml_bldg_id (optional) - If there is more than one <Building> element in an HPXML file,
             use this one. Otherwise just use the first one.
+        hpxml_project_id (optional) - If there is more than one <Project> element in an HPXML file,
+            use this one. Otherwise just use the first one.
```
yzhou601: @nmerket Do we need to consider multiple project and contractor nodes in this case, and connecting each other with ID? This time we are looking up data in three top-level nodes, so maybe it's better to get id recognized?

hescorehpxml/__init__.py
```diff
@@ -735,6 +753,52 @@ def _get_building_address(self, b):
...
 
+    def _get_hpwes(self, b, p, c):
+        xpath = self.xpath
+        hpwes = OrderedDict()
+
+        if p is not None:
+            # need to check if the project is a hpwes program (any other check point?) because "Project" might be used
+            # for other purpose
+            # if not, return blank dict()
+            hpwes_name = ['Home Performance With Energy Star'.upper(), 'HPwES'.upper()]
```
yzhou601: I did this in order to recognize if the project is hpwes project, only in this case we would want to go to the next step.

tests/tests.py
```diff
@@ -1997,6 +1997,74 @@ def test_conditioned_attic(self):
...
+        # contractor ID has to be unique through the whole xml in both places, or error would be given:
+        # "DocumentInvalid: Element '{http://hpxmlonline.com/2014/6}SystemIdentifiersInfo',
+        # attribute 'id': 'c1' is not a valid value of the atomic type 'xs:ID'."
+
+        # At this point, couldn't change Project/ProjectDetails/ContractorSystemIdentifiers/SystemIdentifiersInfo
+        # @id to be the same as
+        # Contractor/ContractorDetails/SystemIdentifier @id
+
+        # Following Building ID logic, it should be done like this:
+        # id under contractor node to be xs:ID type, id under project node could be normal ID type
```
yzhou601: @nmerket please see these comments, test is not passing because some ids have to be unique by its type, even it's under another node. Therefore makes test configuration not successfully validated 🙍

nmerket: I'll have to run these tests to follow what's going on here. Let's talk about this. This may be a misunderstanding of what's an id vs idref or it could be a bug in HPXML itself.

hescorehpxml/__init__.py
```diff
@@ -735,6 +753,52 @@ def _get_building_address(self, b):
...
+            building_bldg_id = xpath(b, 'h:BuildingID/@id')
+            project_contractor_id = xpath(p, '//h:ContractorSystemIdentifier/h:SystemIdentifiersInfo/@id')
+            building_contractor_id = xpath(b, 'h:ContractorID/@id')
+            if project_bldg_id == building_bldg_id:
+                hpwes['building_id'] = building_bldg_id
+            else:
+                return hpwes
+
+            # project information
+            hpwes['improvement_installation_start_date'] = xpath(p, '//h:StartDate/text()')
```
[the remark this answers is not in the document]
yzhou601: Good point, I just checked HEScore API: "The date on which HPwES upgrade installation began, in YYYY-MM-DD format", should be fine!

hescorehpxml/__init__.py
```diff
@@ -735,6 +753,52 @@ def _get_building_address(self, b):
...
+                hpwes['is_income_eligible_program'] = False
+
+            #contractor information
+            if c is not None:
+                contractor_id = xpath(c, 'h:ContractorDetails/h:SystemIdentifier/@id')
+                if contractor_id == building_contractor_id or contractor_id == project_contractor_id:
+                    hpwes['contractor_business_name'] = xpath(c, '//h:BusinessName')
+                    hpwes['contractor_zip_code'] = xpath(c, '//h:BusinessInfo/h:extension/h:ZipCode/text()')
+        # remaining question:
+        # 1. There're two required inputs that I didn't know what they exactly are: "session_token" and "user_key"
```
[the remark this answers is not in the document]
yzhou601: Good to know!

hescorehpxml/__init__.py
```diff
@@ -735,6 +753,52 @@ def _get_building_address(self, b):
...
+        # 2. Error or warning message?
```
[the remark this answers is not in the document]
yzhou601: eg. if these nodes are not pointing to the same building, maybe throw a warning saying that some information in some nodes is ignored? All the other information except for building id is not required so maybe error is not needed.

hescorehpxml/__init__.py
```diff
@@ -735,6 +753,52 @@ def _get_building_address(self, b):
...
+                hpwes['is_income_eligible_program'] = True
+            else:
+                hpwes['is_income_eligible_program'] = False
+
+            #contractor information
+            if c is not None:
+                contractor_id = xpath(c, 'h:ContractorDetails/h:SystemIdentifier/@id')
+                if contractor_id == building_contractor_id or contractor_id == project_contractor_id:
+                    hpwes['contractor_business_name'] = xpath(c, '//h:BusinessName')
+                    hpwes['contractor_zip_code'] = xpath(c, '//h:BusinessInfo/h:extension/h:ZipCode/text()')
```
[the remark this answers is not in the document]
yzhou601: Are we going to propose that to be added to latest hpxml version?

hescorehpxml/__init__.py
```diff
@@ -735,6 +753,52 @@ def _get_building_address(self, b):
...
+            if program_name is None or program_name.text is None:
+                return hpwes
+            elif program_name.text.upper() not in hpwes_name:
+                return hpwes
+
+            # if bldg, project not matching, return blank dict
+            project_bldg_id = xpath(p, 'h:BuildingID/@id')
+            building_bldg_id = xpath(b, 'h:BuildingID/@id')
+            project_contractor_id = xpath(p, '//h:ContractorSystemIdentifier/h:SystemIdentifiersInfo/@id')
+            building_contractor_id = xpath(b, 'h:ContractorID/@id')
```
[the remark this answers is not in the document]
yzhou601: Yes, I did. 
I think the logic is that `Project` is the main node for hpwes program, if `Project` is describing the right bldg, first parse information in `Project`. Then if `Project` connected with `Contractor`, read information from `Contractor`. Then even if there's no contractor information, still some project information could be passed.

hescorehpxml/__init__.py
```diff
@@ -644,7 +648,20 @@ def hpxml_to_hescore_dict(self, hpxml_bldg_id=None, nrel_assumptions=False):
         else:
             b = xpath(self.hpxmldoc, 'h:Building[1]')
 
-        p = xpath(self.hpxmldoc, 'h:Project[1]')
+        if hpxml_project_id is not None:
+            p = xpath(self.hpxmldoc, 'h:Project[h:ProjectID/@id=$projectid]', projectid=hpxml_project_id)
+            if p is None:
+                raise TranslationError('HPXML ProjectID not found')
```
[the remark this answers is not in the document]
yzhou601: 👍

hescorehpxml/__init__.py
yzhou601: 👍 That's exactly what I tried to find!

nmerket:
@yzhou601 Okay, the [Project/ProjectDetails/ContractorSystemIdentifier](https://hpxml.nrel.gov/datadictionary/2.3.0/Project/ProjectDetails/ContractorSystemIdentifiers) is implemented wrong in HPXML. I changed it so we don't use it. I'm also doing less validation that it's the same BuildingID. That whole situation is a mess. Sorry to throw you into it. I think this should be working now.

tests/tests.py
```diff
@@ -61,6 +61,14 @@ def _write_xml_file(self, filename):
     def xpath(self, xpathexpr, *args, **kwargs):
         return self.translator.xpath(self.translator.hpxmldoc, xpathexpr, *args, **kwargs)
 
+    def element_maker(self):
```
yzhou601: Wow this looks cool ! 👍

nmerket:
@yzhou601 can you look at the docs I added and see if they make sense? Once the CI tests are done, the rendered docs should be available in the Artifacts.

docs/source/translation/hpwes.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Inputs for the Home Energy Score `submit_hpwes`_ API call can now be retrieved
+from an HPXML file.
+
+Project
+*******
+
+To get the Home Performance with Energy Star (HPwES) data
+from an HPXML file a ``Project`` node needs to be included with at least the
+following elements:
```
yzhou601: In HEScore API these fields are all optional inputs, it seems here we are asking for the complete inputs. Except for this, other docs look good to me. 😃  
Thanks for cleaning up my codes and simplifying my approach!

PR #103, opened 2019-09-10
allow cooling only mini-split heat pump

### LINKED ISSUE #97: Cooling only mini splits
We are mapping cooling only mini splits to `split_dx` as described [here](https://hescore-hpxml.readthedocs.io/en/latest/translation/cooling_system.html#id2). HEScore has been updated to handle a cooling only mini split and it should be translated as such now. 

- [ ] Code changes
- [ ] Testing
- [ ] Docs

yzhou601:
fix #97

### DIFF
--- a/examples/house2.json
+++ b/examples/house2.json
@@ -1,135 +1,135 @@
 {
   "building_address": {
-    "address": "2 API House", 
-    "city": "Tulsa", 
-    "state": "OK", 
-    "zip_code": "74132", 
+    "address": "2 API House",
+    "city": "Tulsa",
+    "state": "OK",
+    "zip_code": "74132",
     "assessment_type": "initial"
-  }, 
+  },
   "building": {
     "about": {
-      "assessment_date": "2014-10-28", 
-      "shape": "rectangle", 
-      "year_built": 1953, 
-      "number_bedrooms": 3, 
-      "num_floor_above_grade": 2, 
-      "floor_to_ceiling_height": 11, 
-      "conditioned_floor_area": 1620, 
-      "orientation": "east", 
-      "blower_door_test": false, 
+      "assessment_date": "2014-10-28",
+      "shape": "rectangle",
+      "year_built": 1953,
+      "number_bedrooms": 3,
+      "num_floor_above_grade": 2,
+      "floor_to_ceiling_height": 11,
+      "conditioned_floor_area": 1620,
+      "orientation": "east",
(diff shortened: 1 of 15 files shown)


yzhou601:
Test not finished yet

yzhou601:
@nmerket The failed CI test is caused by removal of dummy duct system for mini-split. It compares results by new approach with house4.json, is it supposed to make some changes to `house4.json` to address the new expected results?

hescorehpxml/__init__.py
```diff
@@ -1814,6 +1793,18 @@ def _choose_weighting_factor(systems_dict):
...
+        singleton_heating_systems -= singletons_to_combine
+        singleton_cooling_systems -= singletons_to_combine
+        for heatpump_id in singletons_to_combine:
+            if heating_systems[heatpump_id]['type'] != 'mini_split' or cooling_systems[heatpump_id]['type'] != 'mini_split':  # noqa: E501
+                continue
+            hvac_systems_ids.add(IDsAndWeights(
+                heatpump_id,
+                heatpump_id,
+                None,
+                old_div(heating_systems[heatpump_id][heating_weighting_factor], weight_sum)))
```
yzhou601: I put a separate block for mini-split here instead of treating them as singletons, it is because for small weight mini-split heat pump system (both clg and htg), it's very likely they'll be assigned to different hvac systems if they are entering the following singleton iterations.

nmerket:
@yzhou601 Sorry this took me so long to get to. Yes, do make a change to house4.json if the results _should_ be different now. The python script
https://github.com/NREL/hescore-hpxml/blob/master/hescorehpxml/create_all_example_json.py
will generate new json files that the tests will compare against. You can then diff them to see that the results are changed in the way you expect.

examples/house2.json
```diff
@@ -1,135 +1,135 @@
...
-      "wall_construction_same": false,
-      "window_construction_same": false,
+      ], 
+      "wall_construction_same": false, 
+      "window_construction_same": false, 
       "zone_wall": [
         {
-          "side": "front",
-          "wall_assembly_code": "ewwf00br",
+          "side": "right",
```
yzhou601: I didn't know how these trailing spaces came into being, I diffed regardless of this, changes are only: 1. elimination of dummy distribution system for mini-splits and 2. the zone_wall elements are of different orders.

PR #113, opened 2019-12-10
Windows orientation priority

### LINKED ISSUE #111: Better handle windows attached to foundation walls
If a window is attached to a foundation wall it will try to determine the orientation from the wall but won't be able to find it (It's not looking for foundation walls and they don't have orientation anyway). We should fall back to using the orientation from the window itself in that case.

nmerket:
Fixes #111 

This PR changes the priority of how window orientation is determined to the following:

1. Check for `Azimuth`
2. Check for `Orientation`
3. Use azimuth or orientation of wall it is attached to. 

It also checks to make sure the wall referenced is indeed an above grade wall with an orientation. 

@shorowit, can you review?

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -333,7 +333,7 @@ class TestOtherHouses(unittest.TestCase, ComparatorBase):
         el.getparent().remove(el)
         self.assertRaisesRegexp(
             TranslationError,
-            r'Window\[SystemIdentifier/@id="\w+"\] doesn\'t have Azimuth, Orientation, or AttachedToWall. At least one is required.',
+            r'Window\[SystemIdentifier/@id="\w+"\] doesn\'t have Azimuth, Orientation, or AttachedToWall. At least one is required.',  # noqa E501
             tr.hpxml_to_hescore_dict)
 
     def test_window_only_attached_to_foundation_wall(self):
@@ -346,7 +346,7 @@ class TestOtherHouses(unittest.TestCase, ComparatorBase):
         window.remove(window_orientation)
         self.assertRaisesRegexp(
             TranslationError,
-            r'The Window\[SystemIdentifier/@id="\w+"\] has no Azimuth or Orientation, and the .* didn\'t reference a Wall element.',
+            r'The Window\[SystemIdentifier/@id="\w+"\] has no Azimuth or Orientation, and the .* didn\'t reference a Wall element.',  # noqa E501
             tr.hpxml_to_hescore_dict
         )
 
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -336,6 +336,20 @@ class TestOtherHouses(unittest.TestCase, ComparatorBase):
             r'Window\[SystemIdentifier/@id="\w+"\] doesn\'t have Azimuth, Orientation, or AttachedToWall. At least one is required.',
             tr.hpxml_to_hescore_dict)
 
+    def test_window_only_attached_to_foundation_wall(self):
+        tr = self._load_xmlfile('house4')
+        window_orientation = self.xpath('//h:Window[1]/h:Orientation')
+        window = window_orientation.getparent()
+        etree.SubElement(window, tr.addns('h:AttachedToWall'), attrib={'idref': 'crawlwall'})
+        self._do_compare('house4')
+
+        window.remove(window_orientation)
+        self.assertRaisesRegexp(
+            TranslationError,
+            r'The Window\[SystemIdentifier/@id="\w+"\] has no Azimuth or Orientation, and the .* didn\'t reference a Wall element.',
+            tr.hpxml_to_hescore_dict
+        )
+
     def test_window_attached_to_wall(self):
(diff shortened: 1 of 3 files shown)


nmerket:
Fixes #111

shorowit:
I suggest @yzhou601 review this, she's more familiar with the code.

That said, my reaction to reading the PR description is that it should handle windows on foundation walls. This could happen for a basement dwelling unit, for example.

tests/tests.py
```diff
@@ -333,9 +333,23 @@ def test_missing_window_orientation(self):
...
-            r'All windows need to have either an AttachedToWall, Orientation, or Azimuth sub element\.',
+            r'Window\[SystemIdentifier/@id="\w+"\] doesn\'t have Azimuth, Orientation, or AttachedToWall. At least one is required.',  # noqa E501
             tr.hpxml_to_hescore_dict)
 
+    def test_window_only_attached_to_foundation_wall(self):
+        tr = self._load_xmlfile('house4')
+        window_orientation = self.xpath('//h:Window[1]/h:Orientation')
+        window = window_orientation.getparent()
+        etree.SubElement(window, tr.addns('h:AttachedToWall'), attrib={'idref': 'crawlwall'})
+        self._do_compare('house4')
```
yzhou601: @nmerket I have some concerns probably you've already realized. Or probably I was not understanding this issue correctly.
If I understood this test case correctly, it is: a window is defined as attached to a foundationwall, along with its direction/azimuth information. Executing changes in this branch did follow PR description to prioritize the direction/azimuth information to be used, so the AttachedToWall idref is ignored and results are the same as initial run of house4.xml. 
My concern is, even if this window is successfully passed into HEScore, this test condition passing actually showed it's not modeled as a foundation wall window. It is translated to a child of an above-grade wall of the same side. I didn't know whether it's accepted to directly shift window to be above-grade (or is there any edge case people didn't define an above-grade wall on that side, then this window will be discarded? This edge case executed successfully through our translator, I'm not sure if it will pass HEScore API validation). Assuming this scenario when users are defining windows with idref input pointed to foundation walls, is it indeed expected that this id input is ignored and actually attach it to an above-grade wall? 
I knew the foundation walls are taken by zone_floor to calculate its insulation level instead of a wall input, so it's unavailable to attach a foundation window to a foundation wall. I'm not sure if there's a way to bundle the foundation window + foundation wall into a combined foundation surface then use it to calculate zone_floor insulation? Or is there any better way to model foundation wall windows as inputs? How did people usually do to define a foundation wall window through HEScore API? @lirainer 
At least I think in the documentation or a warning message we should let them know window is modeled as above-grade and the id input is ignored whenever there exists one.

lirainer: In HEScore we do not specify the walls directly - they are imputed from the floor area using an assumed aspect ratio and entered ceiling height and number of stories. For rectangular (non townhouse) homes all four walls are always modeled, so a window attached to a foundation wall should work. The solar modeling will not be accurate, but the thermal model will be okay. At least well within the other assumptions of HEScore.

shorowit: @lirainer Since wall area is imputed and window area is user-entered, is there a concern about having window area > wall area for a given facade?

lirainer: @shorowit Yes, there is an API validation for this and I thought it was checked by hpxml-to-hescore. @nmerket is the area checked?

shorowit: If the hpxml-to-hescore script has a check, it is presumably to ensure that the user-entered window area < user-entered wall area. 

But the API validation sounds more strict, since I presume it checks user-entered window area < imputed wall area. If the API error is hit, **I would think it's more likely that the imputed wall areas are wrong** as opposed to the user-entered window areas being wrong. Maybe it would make more sense to automatically bump up the wall area for a given facade so that it can contain the window area?

shorowit: To be clear, I mean automatically bump up the wall area for the generated E+ (or DOE2) model. Not change anything in hpxml-to-hescore.

lirainer: The hpxml-to-hescore input checks need to mirror the API validation as it is skipping that step and needs to protect the simulation code. Not sure that changing wall areas is a good idea. How much change do you allow? This error happens very rarely and I expect is typically due to input error rather than an odd building.

yzhou601: Thanks @lirainer , if this assumption is working well, then I felt this PR's approach is implemented great! It will be even better if there's somewhere whether in HEScore API or translator docs to clarify how this situation is modeled. Otherwise, like what I said above, somebody doesn't know these insights (e.g. me) might expect window to be modeled along with foundation wall as wall ID specified in HPXML.

nmerket: @lirainer We're not doing the whole geometry calculation in the translator, so there's isn't a validation around window area and wall area.

@yzhou601 There isn't a way to specify a foundation window in HEScore. This at least captures the window area as an above grade window. I suppose the other option would be to ignore foundation windows altogether.

PR #115, opened 2019-12-10
Hvac Weighting Priority

### LINKED ISSUE #110: HVAC weighting priority
Right now the priority for weighting factor on HVAC systems is floor area served then fraction of load. If both have fraction and only cooling (for instance) has floor area, it will throw an error because it wants them to be the same instead of choosing fraction for all of them. This should be changed to pick the "lowest common denominator".

nmerket:
Fixes #110.

### DIFF
--- a/hescorehpxml/__init__.py
+++ b/hescorehpxml/__init__.py
@@ -688,7 +688,7 @@ class HPXMLtoHEScoreTranslator(object):
         bldg['zone']['window_construction_same'] = False
         bldg['zone']['zone_wall'] = self._get_building_zone_wall(b, bldg['about'])
         bldg['systems'] = OrderedDict()
-        bldg['systems']['hvac'] = self._get_hvac(b)
+        bldg['systems']['hvac'] = self._get_hvac(b, bldg['about'])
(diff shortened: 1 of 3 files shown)


hescorehpxml/__init__.py
```diff
@@ -1791,48 +1791,41 @@ def _get_duct_mapping(element_list):
...
-            cooling_weight_sum = 0
-            cooling_weighting_factor = None
 
-        if cooling_systems and heating_systems and heating_weighting_factor != cooling_weighting_factor:
+        # If there's only one heating or cooling system and no weighting elements, fill it in.
+        for heating_or_cooling_systems in (heating_systems, cooling_systems):
+            if len(heating_or_cooling_systems) == 1:
+                for key, hvac_sys in heating_or_cooling_systems.items():
+                    if hvac_sys.get('_floorarea') is None and hvac_sys.get('_fracload') is None:
+                        hvac_sys['_fracloac'] = 1.0
```
lirainer: @nmerket Shouldn't this be using `_fracload`?

nmerket: Good catch!

Issue #120, opened 2020-02-19
HPXML v3 attic roof attachment


yzhou601:
HPXML v3.0 now allows multiple [`AttachedToRoof`](https://hpxml.nrel.gov/datadictionary/3.0.0/Building/BuildingDetails/Enclosure/Attics/Attic/AttachedToRoof) elements in `Attic`. Our current workflow (v2) only allows one-to-one attachment. Should we consider handling that for v3? @nmerket

nmerket:
Maybe try something [like this](https://hescore-hpxml.readthedocs.io/en/latest/translation/zone_roof.html#roof-r-value)

nmerket:
Closed in #121

Issue #91, opened 2019-05-24
Knee Wall Area Accounting


nmerket:
In houses with knee walls, the area of the knee walls should be added to the attic floor area.  If the insulation for the knee walls is different than the attic floor, do a UA calculation to determine the average R-value.   If there is a sloped ceiling area attached to the roof rafters, typical of a cape cod style house, this portion should be characterized as a cathedral ceiling.

@gmdickey

nmerket:
@lirainer Is this going to mess up the footprint calculation checking?

lirainer:
@nmerket Only if we start doing it. We currently do not do any checking of roof and floor areas. @gmdickey any plans?

nmerket:
I think we're good then.

PR #121, opened 2020-02-19
Hpxml v3

yzhou601:
Update translator to be compatible to both hpxml v2 and v3.


hescorehpxml/base.py
```diff
@@ -927,17 +927,10 @@ def get_building_zone_roof(self, b, footprint_area):
                 else:
                     raise TranslationError('If there are more than one Attic elements, each needs an area. If HPXML '
                                            'version 2.*, please specify under Attic/Area, if HPXML version 3.*, '
```
yzhou601: Area used to be read directly in `Attic` element, now it can be specified either in `Roof` being attached or `FrameFloor` being attached, or both. Currently I assume floor area according to conversations [here](https://github.com/NREL/hescore-hpxml/issues/91#issue-448209678). On the other hand, frame floor attachment is not checked, while roof attachment is checked. So there's always a roof element, but might miss floor element. We can either add a floor check or use roof area. Any thought?

hescorehpxml/hpxml3.py
```diff
@@ -0,0 +1,133 @@
...
+        else:
+            raise TranslationError(
+                'Attic {}: Cannot translate HPXML AtticType to HEScore rooftype.'.format(atticid))
+
+    def get_attic_floor_rvalue(self, attic, b):
+        floor_idref = self.xpath(attic, 'h:AttachedToFrameFloor/@idref')
+        frame_floors = self.xpath(b, '//h:FrameFloor[contains(%s, h:SystemIdentifier/@id)]' % floor_idref, aslist=True)
+        rvalue = 0.0
+        for frame_floor in frame_floors:
+            rvalue += self.xpath(frame_floor, 'sum(h:Insulation/h:Layer/h:NominalRValue)')
```
yzhou601: Summing Rvalues of all layers from different frame floors are not making sense. The same question as in https://github.com/NREL/hescore-hpxml/issues/120, for v3.0 where multiple attachments are allowed. Should we consider handling multiple attic roof/floor elements by doing combination calculations? Or document that only one attached element is translated. Maybe I should delete that issue and proceed discussion in this PR?

hescorehpxml/base.py
```diff
@@ -0,0 +1,2245 @@
...
+                    'job completion testing/final inspection': 'final',
+                    'quality assurance/monitoring': 'qa',
+                    'preconstruction': 'preconstruction'
+                }[xpath(b, 'h:ProjectStatus/h:EventType/text()', raise_err=True)]
+            else:
+                assert transaction_type == 'update'
+                bldgaddr['assessment_type'] = 'corrected'
+        return bldgaddr
+
+    def get_hpwes(self, p, c):
```
nmerket: This may not be where you want to use it, but a pattern you could use is the [super function](https://www.pythonforbeginners.com/super/working-python-super-function).

hescorehpxml/base.py
```diff
@@ -0,0 +1,2245 @@
...
+            math.ceil(float(xpath(bldg_cons_el, 'h:NumberofConditionedFloorsAboveGrade/text()', raise_err=True))))
+        avg_ceiling_ht = xpath(bldg_cons_el, 'h:AverageCeilingHeight/text()')
+        if avg_ceiling_ht is None:
+            try:
+                avg_ceiling_ht = float(xpath(bldg_cons_el, 'h:ConditionedBuildingVolume/text()', raise_err=True)) / \
+                                 float(xpath(bldg_cons_el, 'h:ConditionedFloorArea/text()', raise_err=True))
+            except ElementNotFoundError:
+                raise TranslationError(
+                    'Either AverageCeilingHeight or both ConditionedBuildingVolume and ConditionedFloorArea are required.'
+                    # noqa E501
```
nmerket: If you put this on the end of a long line, this will silence the error about the line being too long.

hescorehpxml/base.py
```diff
@@ -0,0 +1,2245 @@
...
+            if atticd['roof_area'] is None:
+                if len(attics) == 1 and len(roofs) == 1:
+                    atticd['roof_area'] = footprint_area
+                else:
+                    raise TranslationError('If there are more than one Attic elements, each needs an area. If HPXML '
+                                           'version 2.*, please specify under Attic/Area, if HPXML version 3.*, '
+                                           'Please specify under the attached frame floor element: FrameFloor/Area.')
+
+            # Roof type
+            self.get_attic_type(attic, atticd, atticid)
```
nmerket: This function has [side effects](https://en.wikipedia.org/wiki/Side_effect_(computer_science)), which means it changes the arguments to the function. It would be better to have it return the value(s) you want to calculate and you save them onto the `atticd` here. It makes it clearer what is going on.

hescorehpxml/base.py
```diff
@@ -0,0 +1,2245 @@
...
+                    ffeffrvalue = floor_eff_rvalues[min(
+                        list(floor_eff_rvalues.keys()),
+                        key=lambda x: abs(ffrvalue - x)
+                    )]
+                    ffua += old_div(ffarea, ffeffrvalue)
+                    fftotalarea += ffarea
+                ffrvalue = old_div(fftotalarea, ffua) - 4.0
+                zone_floor['floor_assembly_code'] = 'efwf%02dca' % min(list(floor_eff_rvalues.keys()),
+                                                                       key=lambda x: abs(ffrvalue - x))
+
```
nmerket: It would be good to move this error catching into a method because we're repeating this a few times.

yzhou601:
#120 Resolved in last two commits

hescorehpxml/base.py
yzhou601: Floor is required for v3, added one test for missing floor error checking

hescorehpxml/hpxml3.py
```diff
@@ -0,0 +1,181 @@
...
+                           aslist=True)
+        return slabs
+
+    def get_foundation_frame_floors(self, fnd, b):
+        attached_ids = self.xpath(fnd, 'h:AttachedToFrameFloor/@idref')
+        frame_floors = self.xpath(b, '//h:FrameFloor[contains("{}",h:SystemIdentifier/@id)]'.format(attached_ids),
+                                  aslist=True)
+        return frame_floors
+
+    def attic_has_rigid_sheathing(self, v2_attic, roof):
```
nmerket, line 52: Note to self: see if there's a cleaner way to do this where the elements are under different parts of the schema by version. It's probably good the way it is.

hescorehpxml/hpxml3.py
```diff
@@ -0,0 +1,181 @@
...
+                         'garage - conditioned': None,  # Fix me
+                         'garage - unconditioned': None,  # Fix me
+                         'garage': 'vented_crawl',
+                         'roof deck': None,  # Fix me
+                         'outside': None,
+                         'attic': None,  # Fix me
+                         'attic - unconditioned': 'uncond_attic',  # Fix me
+                         'attic - conditioned': None,  # Fix me
+                         'attic - unvented': None,  # Fix me
+                         'attic - vented': None}  # Fix me
```
nmerket, line 181: Look through this map and see if we can map more of (or all of?) them.

hescorehpxml/base.py
```diff
@@ -0,0 +1,2331 @@
...
+                    'tinted': 'dtaa',
+                    'solar-control low-e': 'dseaa'
+                }
+            },
+            'Aluminum with Thermal Break': {
+                'double-pane': {
+                    'clear': 'dcab',
+                    'tinted': 'dtab',
+                    'insulating low-e argon': 'dpeaab',
+                    'solar-control low-e': 'dseab'
```
nmerket, line 215: Review this change @nmerket .

nmerket:
@shorowit @lirainer When we specify a roof area in Home Energy Score, is it the area of the thermal boundary (i.e. roof if cathedral ceiling, attic floor if unvented attic) or is it the footprint area of the roof/attic? It's the first, right?

lirainer:
Yes you have it right - it is the first.

docs/source/translation/hvac_distribution.rst
```diff
@@ -35,6 +35,36 @@ following mapping.
    outside                 *not translated*
    ======================  ================
 
+.. table:: Duct Location mapping (HPXML v3)
```
nmerket, line 38: We can do more of the duct mappings. Here are my and @shorowit's suggestions. (Of course, change in the code as well.)

shorowit: In some cases, I suggest a hierarchy where you pick the first HEScore option (space) that exists in the HPXML file.

docs/source/translation/hvac_distribution.rst
```diff
@@ -35,6 +35,36 @@ following mapping.
...
+
+   ===========================  ================
+   HPXML                        HEScore
+   ===========================  ================
+   living space                 cond_space
+   unconditioned space          *not translated*
```
shorowit, line 44:
```suggestion
   unconditioned space          uncond_basement, vented_crawl, unvented_crawl, uncond_attic
```

docs/source/translation/hvac_distribution.rst
```diff
@@ -35,6 +35,36 @@ following mapping.
...
+   under slab                   *not translated*
+   basement                     *not translated*
```
shorowit, line 46:
```suggestion
   basement                     uncond_basement, cond_space
```

docs/source/translation/hvac_distribution.rst
```diff
@@ -35,6 +35,36 @@ following mapping.
...
+   basement - unconditioned     uncond_basement
+   basement - conditioned       cond_space
+   crawlspace - unvented        unvented_crawl
+   crawlspace - vented          vented_crawl
+   crawlspace - unconditioned   *not translated*
```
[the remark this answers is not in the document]
shorowit: Instead of that, I suggest more flexibility:
```suggestion
   crawlspace - unconditioned   vented_crawl, unvented_crawl
```

docs/source/translation/hvac_distribution.rst
```diff
@@ -35,6 +35,36 @@ following mapping.
...
+   crawlspace - conditioned     *not translated*
+   crawlspace                   *not translated*
```
shorowit, line 53:
```suggestion
   crawlspace                   vented_crawl, unvented_crawl, cond_space
```

docs/source/translation/hvac_distribution.rst
```diff
@@ -35,6 +35,36 @@ following mapping.
...
+   exterior wall                *not translated*
+   attic                        *not translated*
```
shorowit, line 55:
```suggestion
   attic                        uncond_attic, cond_space
```

docs/source/translation/hvac_distribution.rst
```diff
@@ -35,6 +35,36 @@ following mapping.
...
+   attic - unconditioned        uncond_attic
+   attic - conditioned          *not translated*
+   attic - unvented             *not translated*
+   attic - vented               *not translated*
+   interstitial space           *not translated*
+   garage                       vented_crawl
```
shorowit, line 61: What happens if the building doesn't have a vented crawl foundation?

docs/source/translation/hvac_distribution.rst
```diff
@@ -35,6 +35,36 @@ following mapping.
...
+   garage - conditioned         *not translated*
+   garage - unconditioned       *not translated*
+   roof deck                    *not translated*
+   outside                      *not translated*
```
shorowit, line 65: If garage -> vented_crawl is ok, then I'd do the same here.
```suggestion
   roof deck                    vented_crawl
   outside                      vented_crawl
```

docs/source/translation/hvac_distribution.rst
```diff
@@ -35,6 +35,36 @@ following mapping.
...
    ======================  ================
 
+.. table:: Duct Location mapping (HPXML v3)
+
+   ===========================  ================
+   HPXML                        HEScore
+   ===========================  ================
+   living space                 cond_space
+   unconditioned space          *not translated*
+   under slab                   *not translated*
```
shorowit, line 45: If garage -> vented_crawl is ok, then I'd do the same here.
```suggestion
   under slab                   vented_crawl
```

docs/source/translation/hvac_distribution.rst
```diff
@@ -35,6 +35,36 @@ following mapping.
    outside                 *not translated*
    ======================  ================
 
+.. table:: Duct Location mapping (HPXML v3)
+
+   ===========================  ================
+   HPXML                        HEScore
```
shorowit, line 41:
```suggestion
   HPXML                        HEScore Hierarchy
```

docs/source/translation/zone_window.rst
```diff
@@ -132,11 +132,38 @@ with storm windows should be considered double-pane.
    other                           *not translated*
    ==============================  ================
 
+.. table:: Window pane mapping for Aluminum frame types (HPXML v3)
```
nmerket, line 135: @lirainer Please review [this translation](https://392-24196988-gh.circle-artifacts.com/0/docs/translation/zone_window.html).

docs/source/translation/zone_window.rst
```diff
@@ -132,11 +132,38 @@ with storm windows should be considered double-pane.
...
+   Starting from HPXML v3, "single-paned with storms" and "single-paned with low-e storms" enumerations
+   are removed. Instead, translator searches ``Window/StormWindow`` element for storm existence.
+   If the storm window is a low-e window, specify ``Window/StormWindow/GlassType`` to be equal to "low-e".
+   ``StormWindow`` is only used when ``single-pane`` window is specified.
+
+   HPXML v2 "single-paned with storms" equivalence(mapped to double-pane) in HPXML v3:
+      - ``Window/GlassLayers`` "single-pane" + ``Window/StormWindow``.
+
+   HPXML v2 "single-paned with low-e storms" equivalence(mapped to double-pane) in HPXML v3:
+      - ``Window/GlassLayers`` "single-pane" + ``Window/StormWindow/StormWindow/GlassType`` to be "low-e".
```
lirainer, line 158:
```suggestion
      - ``Window/GlassLayers`` "single-pane" + ``Window/StormWindow/GlassType`` to be "low-e".
```

docs/source/translation/zone_window.rst
lirainer: @nmerket Translation above looks fine. Found a few things in this section that I think need changing but are not currently part of this PR, so it doesn't appear I can add a comment, so I'll put them here:
* LIne 38: Skylights in HEScore do not have an orientation ~~associated with them~~*that can be set, therefore...

* Line 115: The aluminum frame type ~~has~~ allows for single- and double-paned windows, but not...
* Line 200: low-e                     Double-pane, ~~clear~~**solar-control low-E**
* Line 528: For each side of the house in HEScore, ~~a~~ solar screen**s** may be present. 
* Line 529: To determine if ~~a~~ solar screen**s** should be specified, the translator looks for either

docs/source/translation/zone_window.rst
yzhou601: Good catch. Thanks, will address it.

lirainer: Forgot to mention that this same error occurs in a least a couple more sections below.

docs/source/translation/zone_window.rst
yzhou601: @lirainer These all turn out to be doc issues. The third one is an [improvement](https://github.com/NREL/hescore-hpxml/pull/100/files) @nmerket did, maybe forgot to update docs, we can easily address these in the current PR.

tests/tests.py
```diff
@@ -2915,6 +2925,37 @@ def test_townhouse_walls_translation(self):
...
+        self.assertEqual(d_v3['building']['systems']['hvac'][0]['hvac_distribution'][0]['location'], 'cond_space')
+
+        el.text = 'attic'
+        d_v3 = tr_v3.hpxml_to_hescore()
+        self.assertEqual(d_v3['building']['systems']['hvac'][0]['hvac_distribution'][0]['location'], 'uncond_attic')
+
+        # Is this reasonable?
+        el.text = 'crawlspace'
+        d_v3 = tr_v3.hpxml_to_hescore()
+        self.assertEqual(d_v3['building']['systems']['hvac'][0]['hvac_distribution'][0]['location'], 'cond_space')
```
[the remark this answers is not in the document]
yzhou601: Right, it's not easy. If we want to do the space error-checking based on HPXML (before translation), it's obviously more strict and complicated regarding to facts that there're more enumerations, the enumerations of the same space may be slightly different in different HPXML elements, schema and our mappings are changing, etc. And it's also hard to know where it should be placed along translation. Because we are discarding some extra elements during translation (eg. only two roofs are retained), and we want to make sure the final translated elements are in good shape, so it's reasonable to put it at the end.  But sometimes it's not easy to know what is the mapping between elements in HPXML and elements in translated HEScore inputs, so retrieving information from HPXML at the end may be hard. 
I am not sure if it worth the time and effort to improve this error-checking, so I leave it there for discussion.

nmerket:
- [ ] @nmerket to look at HPXML v3 test files to ensure they make sense.

nmerket:
I did a cursory look at the HPXML files you created, @yzhou601, and they look good. I'm gonna merge this.

PR #123, opened 2020-07-01
Fixing CLI

nmerket:
For some reason, either in python3 or a newer version of lxml, the way we were reading in files from the command line no longer was able to be parsed by lxml. This was not caught by the tests because there were no tests on the CLI. This fixes the original problem _and_ adds a test for the CLI so we'll know if it regresses again.

cc @adamsten

### DIFF
--- a/hescorehpxml/__init__.py
+++ b/hescorehpxml/__init__.py
@@ -18,11 +18,10 @@ def HPXMLtoHEScoreTranslator(hpxmlfilename):
         raise HPXMLtoHEScoreError('Schema version {} not supported.'.format('.'.join(schema_version)))
 
 
-def main():
+def main(argv=sys.argv[1:]):
     parser = argparse.ArgumentParser(description='Convert HPXML v2.x or v3.x files to HEScore inputs')
     parser.add_argument(
         'hpxml_input',
-        type=argparse.FileType('r'),
         help='Filename of hpxml file'
     )
     parser.add_argument(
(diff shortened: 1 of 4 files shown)


hescorehpxml/__init__.py
```diff
@@ -18,11 +18,10 @@ def HPXMLtoHEScoreTranslator(hpxmlfilename):
         raise HPXMLtoHEScoreError('Schema version {} not supported.'.format('.'.join(schema_version)))
 
 
-def main():
+def main(argv=sys.argv[1:]):
     parser = argparse.ArgumentParser(description='Convert HPXML v2.x or v3.x files to HEScore inputs')
     parser.add_argument(
         'hpxml_input',
-        type=argparse.FileType('r'),
```
nmerket, line 25: This is the actual fix. For whatever reason this is returning a [TextIOWrapper](https://docs.python.org/3/library/io.html#io.TextIOWrapper), which wasn't being read by the lxml.etree.parse function throwing an error on every file. I changed this to be a string so it will just pass the path and it works now. 

All the rest of the changes in this PR are adding testing to the CLI code so this kind of error will be caught in the future.

tests/tests.py
```diff
@@ -104,6 +105,20 @@ def test_house8(self):
         self._do_full_compare('house8')
 
 
+class TestCLI(unittest.TestCase, ComparatorBase):
```
yzhou601, line 108: Good test that captures input and output ports between translator and CLI.  I didn't realize we're testing all the features by directly calling class functions, now it's absolutely more complete. 👍

PR #127, opened 2020-10-06
Add export of HPXML file scrubbed of PII

### LINKED ISSUE #125: Export an HPXML file scrubbed of PII
It would be helpful to be able have the HPXML file a user submits for support and debugging. However, we want to remove any personally identifiable information, so let's add a flag that optionally writes the imported HPXML file with the following elements removed

- All data under “Customer”
- Under “Building/Site/Address”, All data with “Address Type” = “Mailing” should be removed (I am assuming the “Street” address type is where we get that information for HEScore)
- All data under “Health and Safety”
- All data under “Building/Building Details/Building Summary/Building Occupancy”
- All data under “Building Summary/Annual Energy Use”
- All data under “Utility”
- All data under “Consumption”
- All data under “Building/CustomerID”.

cc @lirainer @gmdickey

nmerket:
Fixes #125.

This PR does the following:
- Adds a cli arg `--scrubbed-hpxml` to optionally export an HPXML file with PII removed.
- Switches the testing framework to pytest.
- Removes compatibility with Python 2.7. We were getting a lot of deprecation warnings in the testing. Python 2.7 [reached end of life on Jan 1, 2020](https://www.python.org/doc/sunset-python-2/), so it seems like this is an appropriate time to drop support.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,9 @@ jobs:
           name: Run Unittests
           command: |
             source env/bin/activate
-            pytest
+            pytest --junitxml=/tmp/test_results
+      - store_test_results:
+          path: /tmp/test_results
       - run:  &stylechecks
           name: Run Style Checks
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
           name: Run Unittests
           command: |
             source env/bin/activate
-            pytest --junitxml=/tmp/test_results
+            pytest --junitxml=/tmp/test_results/hescore-hpxml/results.xml
       - store_test_results:
           path: /tmp/test_results
       - run:  &stylechecks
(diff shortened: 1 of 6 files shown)


lirainer:
@nmerket In #125 it has:
> `Building/Site/Address`, All data with `Address Type` = `Mailing` should be removed

But I don't see anything related to Building/Site in export_scrubbed_hpxml

nmerket:
@lirainer I omitted that because:
1. There is only one `Building/Site/Address` element allowed.
2. We [require it to be a street address](https://github.com/NREL/hescore-hpxml/blob/d46f8002505438d16abbb397cd7bb23f3e279c6a/hescorehpxml/base.py#L737), so if the house is submitted with a mailing address there the translation would've failed and never got to this point. 
3. If I remove the address, the file will fail to validate. I suppose I could just remove all the fields except zip code.

lirainer (approved):
Looks good to me. I tested it on my Mac which is still running 2.7 and I didn't get any errors, which is surprising. Also ran it in a 3.6 environment with no issues.

nmerket:
Yeah, I think the incompatible stuff is in the testing suite.

PR #128, opened 2020-10-21
Exporting scrubbed file even when the translation fails

nmerket:


lirainer (commented):
I guess I'm not quite clear on the goal of this. It looks to me like all of the exceptions lead to a sys.exit() and will not reach `export_scrubbed_hpxml`. Also, it would be good to distinguish the errors returned by a `HPXMLtoHEScoreTranslator` exception from a `hpxml_to_hescore_json` exception. Right now they are identical.

nmerket:
The first `try-except` will error if the file doesn't validate against the schema. I should probably return those validation errors. That might be helpful to users. I don't want to export the scrubbed HPXML in that case because I don't know that the file is valid HPXML and if/where any PII is in that case.

The second one happens if there is a translation fails due to missing required or inconsistent elements. In that case the `finally` block runs afterwards no matter whether there was an exception or not.

lirainer (approved):
Looks good to me. Verified that it outputs the scrubbed file when there is a translation error, but not when there is a schema error.

PR #131, opened 2020-12-16
Area weighted R-value fix for knee walls

### LINKED ISSUE #129: Area weighted R-value fix for knee walls
Proposed changes:

- Get rid of the center of cavity effective R-value lookups for walls and attics. Do area weighted R-values with nominal cavity R-values only.
- If the continuous insulation is >= R-5, choose the appropriate construction type. Do not add additional continuous R-value to the cavity R-value (not sure if we're already doing this).
- For cases where one surface is close to R-0, set it to some small minimum R-value. @lirainer is going to come up with a very well thought out value to use here.

@lirainer @gdickey627

bpark1327:
Closes #129

### DIFF
--- a/examples/house8.json
+++ b/examples/house8.json
@@ -27,7 +27,7 @@
           "roof_assembly_code": "rfwf00tg",
           "roof_color": "medium",
           "roof_type": "vented_attic",
-          "ceiling_assembly_code": "ecwf30",
+          "ceiling_assembly_code": "ecwf38",
           "zone_skylight": {
             "skylight_area": 0
           }
@@ -63,26 +63,6 @@
       "window_construction_same": false,
       "zone_wall": [
         {
-          "side": "back",
-          "wall_assembly_code": "ewwf00wo",
-          "zone_window": {
-            "window_area": 220.0,
-            "window_method": "code",
-            "window_code": "dpeaw",
-            "solar_screen": false
-          }
-        },
-        {
-          "side": "right",
-          "wall_assembly_code": "ewwf00wo",
-          "zone_window": {
-            "window_area": 40.0,
-            "window_method": "code",
-            "window_code": "dpeaw",
-            "solar_screen": false
-          }
-        },
-        {
           "side": "front",
           "wall_assembly_code": "ewwf00wo",
           "zone_window": {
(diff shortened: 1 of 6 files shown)


bpark1327:
Insulated knee wall:
image

Uninsulated knee wall:
image
Note that an arbitrary high UA value is used for R-0 knee wall.

hescorehpxml/hpxml3.py
```diff
@@ -60,6 +60,32 @@ def get_attic_roof_rvalue(self, v2_attic, roof):
         return self.xpath(roof,
                           'sum(h:Insulation/h:Layer/h:NominalRValue)')
 
+    def get_attic_knee_wall_rvalue_and_area(self, attic, b):
```
nmerket, line 63: This function is identical to the one in hpxml2.py. Since it's the same in both places, it should be removed from both, placed in base.py instead so it can be inherited by the sub classes.

tests/test_translation.py
```diff
@@ -928,7 +928,7 @@ def test_attic_knee_wall_zero_rvalue(self):
         kneewall_rvalue_el.text = '0'
         # run translation
         resp = tr.hpxml_to_hescore()
-        self.assertEqual(resp['building']['zone']['zone_roof'][0]['ceiling_assembly_code'], 'ecwf00')
+        self.assertEqual(resp['building']['zone']['zone_roof'][0]['ceiling_assembly_code'], 'ecwf03')
```
nmerket, line 931: Looks to be working as intended.

examples/house8.json
```diff
@@ -27,7 +27,7 @@
           "roof_assembly_code": "rfwf00tg",
           "roof_color": "medium",
           "roof_type": "vented_attic",
-          "ceiling_assembly_code": "ecwf30",
+          "ceiling_assembly_code": "ecwf38",
```
nmerket, line 30: Interesting that this one changed as it doesn't have knee walls. It doesn't seem like that should be the case. Isn't the calculation staying the same for buildings that don't have knee walls?

hescorehpxml/base.py
```diff
@@ -1153,20 +1153,16 @@ def get_building_zone_roof(self, b, footprint_area):
...
+
+                attic_floor_coc_rvalue = attic_floor_rvalue + 0.5
+                attic_floor_adj_ua = knee_wall_ua + atticd['roof_area'] / attic_floor_coc_rvalue
                 attic_floor_adj_area = knee_wall_area + atticd['roof_area']
-                attic_floor_rvalue = attic_floor_adj_area / attic_floor_adj_ua
+                attic_floor_rvalue = (attic_floor_adj_area / attic_floor_adj_ua) - 0.5
                 atticd['roof_area'] = attic_floor_adj_area
-            atticd['attic_floor_coc_rvalue'] = \
-                min(attic_floor_rvalues, key=lambda x: abs(x - attic_floor_rvalue)) + 0.5
+            atticd['attic_floor_coc_rvalue'] = attic_floor_rvalue + 0.5
```
nmerket, line 1165: Oh, right there was a slight difference in the calculation because we're not snapping it to one of the construction code values here, so there could be a difference. So, that house8 thing is probably fine.

nmerket (changes requested):
Looks pretty good to me. Just the one change about where to put that `get_attic_knee_wall_rvalue_and_area` method.

hescorehpxml/base.py
bpark1327: That's right. `house8` was changed because of this.

examples/house8.json
bpark1327: The next comment explains this change.

nmerket (approved):

nmerket:
Thanks @bpark1327!

PR #137, opened 2021-04-13
Fix cap of number of bedrooms

### LINKED ISSUE #136: NumberofBedrooms silently capped at 10
If `NumberofBedrooms` is greater than 10 then hpxml2hescore returns `number_bedrooms` set to 10 and no error.

Test file: 
[bad_input_hpxml.txt](https://github.com/NREL/hescore-hpxml/files/6225296/bad_input_hpxml.txt)
* Currently: returns json with `number_bedrooms` set to 10
* Expected: returns InputOutOfBounds error

yzhou601:
fix #136

### DIFF
--- a/examples/house6.xml
+++ b/examples/house6.xml
@@ -32,7 +32,7 @@
                     <ResidentialFacilityType>single-family detached</ResidentialFacilityType>
                     <NumberofConditionedFloorsAboveGrade>2</NumberofConditionedFloorsAboveGrade>
                     <AverageCeilingHeight>8</AverageCeilingHeight>
-                    <NumberofBedrooms>11</NumberofBedrooms>
+                    <NumberofBedrooms>10</NumberofBedrooms>
                     <ConditionedFloorArea>2600</ConditionedFloorArea>
                 </BuildingConstruction>
             </BuildingSummary>
--- a/tests/test_translation.py
+++ b/tests/test_translation.py
@@ -1149,6 +1149,14 @@ class TestInputOutOfBounds(unittest.TestCase, ComparatorBase):
                                'num_floor_above_grade is out of bounds',
                                tr.hpxml_to_hescore)
 
+    def test_number_bedrooms(self):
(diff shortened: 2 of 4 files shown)


examples/house6.xml
```diff
@@ -32,7 +32,7 @@
                     <ResidentialFacilityType>single-family detached</ResidentialFacilityType>
                     <NumberofConditionedFloorsAboveGrade>2</NumberofConditionedFloorsAboveGrade>
                     <AverageCeilingHeight>8</AverageCeilingHeight>
-                    <NumberofBedrooms>11</NumberofBedrooms>
```
yzhou601, line 35: Example house6 has 11 bedrooms which is out of bound, I changed it to 10 to make our tests slip through but not sure whether it is an acceptable solution. @nmerket @lirainer

tests/test_translation.py
```diff
@@ -1149,6 +1149,14 @@ def test_num_floor_above_grade(self):
                                'num_floor_above_grade is out of bounds',
                                tr.hpxml_to_hescore)
 
+    def test_number_bedrooms(self):
+        tr = self._load_xmlfile('hescore_min')
+        el = self.xpath('//h:NumberofBedrooms')
+        el.text = '25'
+        self.assertRaisesRegex(InputOutOfBounds,
+                               'number_bedrooms is out of bounds',
+                               tr.hpxml_to_hescore)
```
nmerket, line 1158: Are we still hitting this?

examples/house6.xml
nmerket: These test houses come from some descriptions of houses Glenn gives to software developers to test how they're using the API. I'm pretty sure we had 11 here to see if it dropped it to 10.

nmerket (commented):
I'm pretty sure this and the "problem" in #136 isn't a bug, but a feature and was intentional at one point. Let's talk with Glenn on Thursday about this and see if we can remember why we did this in the first place.

nmerket (approved):

PR #140, opened 2021-04-14
Changing knee wall R-value adder to R-1.8

nmerket:
As requested 

@gamalielL @gdickey627

### DIFF
--- a/docs/source/translation/zone_roof.rst
+++ b/docs/source/translation/zone_roof.rst
@@ -360,7 +360,7 @@ wall". See below an example:
    <Walls>
 
 If an attic has knee walls specified, the area of the knee walls will be added
-to the attic floor area. The knee walls center-of-cavity R-value is R-0.5 greater 
+to the attic floor area. The knee walls center-of-cavity R-value is R-1.8 greater 
 than the nominal R-value. The knee walls center-of-cavity R-value will be reflected 
 in the area weighted center-of-cavity effective R-value of the attic floor. 
 This averaged center-of-cavity effective R value is combined from all knee walls 
--- a/hescorehpxml/base.py
+++ b/hescorehpxml/base.py
@@ -1180,7 +1180,7 @@ class HPXMLtoHEScoreTranslatorBase(object):
             attic_floor_rvalue = self.get_attic_floor_rvalue(attic, b)
             if knee_walls:
                 knee_wall_rvalue, knee_wall_area = self.get_attic_knee_wall_rvalue_and_area(attic, b)
-                knee_wall_coc_rvalue = knee_wall_rvalue + 0.5
+                knee_wall_coc_rvalue = knee_wall_rvalue + 1.8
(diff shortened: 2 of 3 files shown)


bpark1327 (approved):
Looks good to me. 

Insulated knee wall:
image

Uninsulated knee wall:
image
Please note that an arbitrary high UA value is used for R-0 knee wall in original calculation.

GamalielL:
The attic floor should also have R-1.8 added for the coc R-value. Aside
from the differences between horizontal and vertical air films, the knee
wall and attic surfaces are the same.  The issue with 0.5 is that it gives
too much weight to uninsulated surfaces and that will still be a problem
when you have to combine two attic surfaces.  It also doesn't make sense to
add 1.8 to the knee wall, but then subtract 0.5 from the combined
assembly.

Gamaliel Lodge
he/him/his
OptiMiser LLC
www.optimiserenergy.com

nmerket:
@GamalielL I'm the first to admit that the _workaround_ in this tool for combining multiple attics (and walls) is incorrect. I've penciled it out and it is not correctly applying a parallel path calculation, which should be done for the entire wall/attic section and then an equivalent R-value backed out. My understanding is that it is a rough approximation originally given to assessors as a back of the envelope calculation they could do when they had more wall or attic sections than the user interface would allow them to enter. In this translator we tried to implement that for consistency with that direction given to the assessors. @gdickey627 could probably explain the history of that better than I could. 

This inaccuracy is further compounded by the additional add-on workaround to include knee walls in the attic floor area using an additional simplified (but still wrong) parallel path calculation. As we know, this really starts to fall apart at R-values close to zero. This PR attempts to put a band-aid on a band-aid on a band-id, but it does give moderately more reasonable results. 

At this point I'm done trying to defend it or fix it. As I see it, here's the way we get out of this mess:

1. Add inputs to the Home Energy Score API for knee walls and model them as such. No more of this combining with the attic nonsense.
2. Remove the arbitrary limit on the number of attic/wall/etc sections an assessor can input. Move these parallel path calculations into the modeling software where they should be.

Next week we're releasing to the public the OpenStudio/EnergyPlus version of Home Energy Score. Changes such as these have been off the table for a while since we didn't want to invest making these modeling improvements in the old DOE2.1e model articulation code that we were going to throw away soon anyway. After this release we will have a more flexible and maintainable modeling backend and can begin to work on things that will change the Home Energy Score API. The items above are high on my list for the very next version.

As for this, I can merge it, or not. Let me know. I really don't have a preference. With any luck we can delete this whole section of code in a few months and never speak of it again.

GamalielL:
Noel,

I wasn't trying to be hard on you with that last comment, just musing. I
agree with your assessment and approach. I'm glad the move to EPlus will
make things easier on your end.  I do think we should merge that change for
the Knee Walls, since we have a customer waiting on it.

Thanks for all your hard work!

Gamaliel Lodge
he/him/his
OptiMiser LLC
www.optimiserenergy.com

PR #144, opened 2021-05-27
JSON Schema

### LINKED ISSUE #143: Add json schema validation to hescore-hpxml

bpark1327:
closes #143


hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2208 @@
...
+                        },
+                        "blower_door_test": {
+                            "$id": "#/properties/building/properties/about/properties/blower_door_test",
+                            "type": "boolean",
+                            "description": "Was a blower door test performed on this house?"
+                        },
+                        "air_sealing_present": {
+                            "$id": "#/properties/building/properties/about/properties/air_sealing_present",
+                            "type": "boolean",
+                            "description": "Has the building been air sealed? (only used if blower_door_test is 0 )"
```
shorowit, line 141: I assume this should say false instead of 0.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2208 @@
...
+                        },
+                        "envelope_leakage": {
+                            "$id": "#/properties/building/properties/about/properties/envelope_leakage",
+                            "type": "integer",
+                            "minimum": 0,
+                            "maximum": 25000,
+                            "description": "Building air leakage measured with a blower door (CFM50) (only used if blower_door_test is 1 )"
```
shorowit, line 148: I assume this should say true instead of 1.

nmerket:
Tests are failing because #147.

bpark1327:
> Tests are failing because #147.

@nmerket I think what we need to do to resolve CI test failures due to JSON schema validation errors includes:
- [ ] Move [the `external_building_id` code](https://github.com/NREL/hescore-hpxml/blob/bc87aac9d75c7266ca4fd04e43060f54854bc326/hescorehpxml/base.py#L879-L887)  to `get_building_address`
- [ ] In test_external_id_passthru and test_external_id_extension_passthru, change `self.assertEqual(myid, res['building']['about'].get('external_building_id'))` to `self.assertEqual(myid, res['building_address'].get('external_building_id'))`

Would you mind if I address it here? Or would it make more sense to open another PR for this?

nmerket:
@bpark1327 Let's open a separate PR for #147. Then it can be a small, independent change that we can merge back into master, which should fix this PR. Wanna take a crack at it?

bpark1327:
> @bpark1327 Let's open a separate PR for #147. Then it can be a small, independent change that we can merge back into master, which should fix this PR. Wanna take a crack at it?

@nmerket Sounds good. I will take a crack at it.

bpark1327:
> > @bpark1327 Let's open a separate PR for #147. Then it can be a small, independent change that we can merge back into master, which should fix this PR. Wanna take a crack at it?
> 
> @nmerket Sounds good. I will take a crack at it.

Okay. It is being addressed in #149.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2662 @@
...
+                                        }
+                                    ]
+                                }
+                            }
+                        }
+                    }
+                }
+            },
+            "additionalProperties": false,
+            "allOf": [
```
bpark1327, line 1892: Note that nested properties in JSON schema cannot see up the tree. Therefore, I put constraints for walls and windows at `building` level because different set of constraints need to be applied based on properties that are at upper level of the tree. These properties include `building.about.shape`, `building.zone.wall_construction_same`, and `building.zone.window_construction_same`.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2662 @@
...
+                                                "$id": "#/properties/building/properties/zone/properties/zone_roof/items/properties/zone_skylight/properties/solar_screen",
+                                                "type": "boolean",
+                                                "description": "Does this window have a solar screen?"
+                                            }
+                                        },
+                                        "allOf": [
+                                            {
+                                                "if": {
+                                                    "properties": {
+                                                        "skylight_area": {"not": {"const": 0}}
```
bpark1327, line 356: What I need to check here is whether `skylight_area` is greater than 0. However, as far as I know, it cannot be done with plain JSON schema. There doesn't seem to be no such a thing as `greater than`. Thus, I used a combination of min/max and "not zero" checking.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2662 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+            },
+            "additionalProperties": false,
+            "allOf": [
+                {
+                    "if": {
```
bpark1327, line 1894: There seems to be a way to simplify this part, but I couldn't think of it. I am open to any suggestions.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2662 @@
...
+    "additionalProperties": false,
+    "definitions": {
+        "def_front_wall_constraint": {
+            "contains": {
+                "properties": {
+                    "side": {"enum": ["front"]}
+                },
+                "required": ["side", "wall_assembly_code"]
+            },
+            "error_msg": "zone_wall/side[\"front\"] requires \"side\" and \"wall_assembly_code\""
```
bpark1327, line 2240: Here I use "contains" method to check whether the list of zone_wall contains a valid front wall property (i.e., the front wall with `side` and `wall_assembly_code`). If a json file doesn't meet this condition, "contains" method will generate an incomprehensible error message. So I use custom message here.

nmerket:
Note to self: check for UEF.

hescorehpxml/schemas/hescore_json.schema.json
nmerket: I think [exclusiveMinimum](https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.2.5) is what you're looking for.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2482 @@
...
+                                        },
+                                        "then": {
+                                            "required": ["floor_area"]
+                                        }
+                                    },
+                                    {
+                                        "if": {
+                                            "properties": {
+                                                "floor_name": {"enum": ["floor1"]},
+                                                "floor_area": {"not": {"const": 0}}
```
nmerket, line 510: Same about the exclusiveMinimum.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2482 @@
...
+                                                 "ewov21wo", "ewov21st", "ewov21vi", "ewov21al", "ewov21br",
+                                                 "ewov27wo", "ewov27st", "ewov27vi", "ewov27al", "ewov27br",
+                                                 "ewov33wo", "ewov33st", "ewov33vi", "ewov33al", "ewov33br",
+                                                 "ewov38wo", "ewov38st", "ewov38vi", "ewov38al", "ewov38br",
+                                                 "ewbr00nn", "ewbr05nn", "ewbr10nn",
+                                                 "ewcb00st", "ewcb00br", "ewcb00nn",
+                                                 "ewcb03st", "ewcb03br", "ewcb03nn",
+                                                 "ewcb06st", "ewcb06br", "ewcb06nn",
+                                                 "ewsb00st"],
+                                        "maxLength": 20,
```
nmerket, line 589: Are these maxLength specifiers necessary when you have enums?

hescorehpxml/schemas/hescore_json.schema.json
nmerket: Sounds good. Glad you figured this out.

hescorehpxml/schemas/hescore_json.schema.json
nmerket: I don't know. It might just be complicated. I'm not seeing it either.

hescorehpxml/schemas/hescore_json.schema.json
nmerket: Nice use of custom error message! 💯

tests/test_schema.py
```diff
@@ -0,0 +1,514 @@
...
+            errors.append(error.schema["error_msg"])
+        except KeyError:
+            errors.append(error.message)
+    return errors
+
+
+def test_schema_version_validation():
+    schema = get_json_schema()
+    error = jsonschema.Draft7Validator.check_schema(schema)
+    assert error is None
```
nmerket, line 44: I like that you have a test to see if the schema itself is valid.

tests/test_schema.py
```diff
@@ -0,0 +1,514 @@
...
+import json
+import pathlib
+import copy
+import glob
+import pytest
+
+hescore_examples = [
+    'townhouse_walls',
+    'house1'
+]
```
nmerket, line 12: Nice test suite. I liked the use of parametrization in pytest. Do we need to have more of the examples here? Or is it covered in existing tests because the json outputs of all the existing tests are validated against this schema?

nmerket (commented):
This looks great! One or two minor comments below.

lirainer:
I'm writing a JSON schema and this has been a big help! A couple of comments:
* `skylight_shgc` and `window_shgc` should be using exclusiveMinimum and exclusiveMaximum
*  Use of `const` vs `enum` with a single value (such as [here](https://github.com/NREL/hescore-hpxml/blob/fe286226e6fe68e83e7986da4763861d6f5faa55/hescorehpxml/schemas/hescore_json.schema.json#L1558)) is inconsistent. Probably should use just `const` unless there is a good reason?

hescorehpxml/schemas/hescore_json.schema.json
bpark1327: Indeed, these are not necessary. Will remove them.

bpark1327:
> I'm writing a JSON schema and this has been a big help! A couple of comments:
> 
> * `skylight_shgc` and `window_shgc` should be using exclusiveMinimum and exclusiveMaximum
> * Use of `const` vs `enum` with a single value (such as [here](https://github.com/NREL/hescore-hpxml/blob/fe286226e6fe68e83e7986da4763861d6f5faa55/hescorehpxml/schemas/hescore_json.schema.json#L1558)) is inconsistent. Probably should use just `const` unless there is a good reason?

@lirainer Thanks for your comments. As to your second comment, I agree that I should use `const` for a single value.

tests/test_schema.py
bpark1327: I think the existing tests cover most of what we want to test. But if you think we need other examples, please let me know.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2452 @@
...
+            }
+        },
+        "hpwes": {
+            "$id": "#/properties/hpwes",
+            "type": "object",
+            "properties": {
+                "improvement_installation_start_date": {
+                    "$id": "#/properties/building/properties/improvement_installation_start_date",
+                    "type": "string",
+                    "pattern": "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$",
```
nmerket, line 46: This could probably replaced with the format="date" if you're using draft 7. See https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times

nmerket (commented):
- [ ] Change it where it does an either or (e.g. Windows, Skylights, HVAC, Water Heating, PV) to enforce that in the schema. Talked to @lirainer and we're good to go with this.

hescorehpxml/schemas/hescore_json.schema.json
bpark1327: Lesson learned: to validate `date` format, I need to specify the `format_checker` argument when calling `validate` or `Draft7Validator`.

nmerket: Ok. Sounds good. If it's a real problem the current regex implementation works too. It's just that using the `date` format would be more semantically clear. It's a pretty minor issue.

lirainer:
> * [ ]  Change it where it does an either or (e.g. Windows, Skylights, HVAC, Water Heating, PV) to enforce that in the schema. Talked to @lirainer and we're good to go with this.

@nmerket Does this extend to **all** elements which provide a choice? In addition to the elements we discussed, such as `zone_skylight.skylight_method` or `heating.efficiency_method`, there are many, many others that influence what other elements are used:
* about.shape
* about.town_house_walls
* about.blower_door_test
* zone.wall_construction_same
* zone.window_construction_same
* zone_roof.roof_type
* zone_roof.roof_color
* zone_floor.foundation_type
...

Implementing all of these decisions in the JSON file creation method would be essentially recreating the validation step, which would have already been run.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2452 @@
...
+                            "type": "array",
+                            "additionalItems": false,
+                            "minItems": 0,
+                            "maxItems": 2,
+                            "items": {
+                                "$id": "#/properties/building/properties/systems/properties/hvac/items/",
+                                "type": "object",
+                                "required": [
+                                    "hvac_name"
+                                ],
```
nmerket, line 640: Again, I know more than this is required. It looks like some of the requirements may be specified later, but I wonder if it could be simplified.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2452 @@
...
+                                "fuel_primary": {
+                                    "$id": "#/properties/building/properties/systems/properties/domestic_hot_water/properties/fuel_primary",
+                                    "type": "string",
+                                    "enum": ["electric", "natural_gas", "lpg", "fuel_oil"],
+                                    "description": "Primary water heater fuel"
+                                },
+                                "efficiency_method": {
+                                    "$id": "#/properties/building/properties/systems/properties/domestic_hot_water/properties/efficiency_method",
+                                    "type": "string",
+                                    "enum": ["user", "uef", "shipment_weighted"],
```
nmerket, line 1425: UEF ✅

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2452 @@
...
+                                            "properties": {
+                                                "heating": {
+                                                    "properties": {
+                                                        "type": {"const": "boiler"}
+                                                    },
+                                                    "required": ["type"]
+                                                }
+                                            }
+                                        },
+                                        "error_msg": "The category element can only be set to \"combined\" if the heating/type is \"boiler\" and the boiler provides the domestic hot water"
```
nmerket, line 1673: Nice use of intelligible error messages.

nmerket (commented):
This is really coming along. I look forward to the additional validation around xor logic with windows, etc.

nmerket:
> > * [ ]   Change it where it does an either or (e.g. Windows, Skylights, HVAC, Water Heating, PV) to enforce that in the schema. Talked to @lirainer and we're good to go with this.
> 
> @nmerket Does this extend to **all** elements which provide a choice? In addition to the elements we discussed, such as `zone_skylight.skylight_method` or `heating.efficiency_method`, there are many, many others that influence what other elements are used:
> 
>     * about.shape
> 
>     * about.town_house_walls
> 
>     * about.blower_door_test
> 
>     * zone.wall_construction_same
> 
>     * zone.window_construction_same
> 
>     * zone_roof.roof_type
> 
>     * zone_roof.roof_color
> 
>     * zone_floor.foundation_type
>       ...
> 
> 
> Implementing all of these decisions in the JSON file creation method would be essentially recreating the validation step, which would have already been run.

@lirainer I think that's fine to recreate the validation here. That gives us some surety on the inputs. Maybe I'm misunderstanding your concerns.

hescorehpxml/schemas/hescore_json.schema.json
bpark1327: It's not a problem. I prefer to use `date` format. I'll make this change.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2452 @@
...
+                    "type": "object",
+                    "required": [
+                        "wall_construction_same",
+                        "window_construction_same"
+                    ],
+                    "properties": {
+                        "zone_roof": {
+                            "$id": "#/properties/building/properties/zone/properties/zone_roof",
+                            "type": "array",
+                            "maxLength": 20,
```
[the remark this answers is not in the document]
bpark1327: You're absolutely right. This is meant for strings. I'll remove it.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2452 @@
...
+                        "comments": {
+                            "$id": "#/properties/building/properties/about/properties/comments",
+                            "type": ["string", "null"],
+                            "maxLength": 256,
+                            "description": "Free text comment field"
+                        },
+                        "comment_api_only": {
+                            "$id": "#/properties/building/properties/about/properties/comment_api_only",
+                            "type": "string",
+                            "maxLength": 40,
```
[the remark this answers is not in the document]
bpark1327: Should we remove `maxLength` for string properties?

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2452 @@
...
+                            "type": "array",
+                            "description": "Inputs about the construction of the floor(s)",
+                            "additionalItems": false,
+                            "minItems": 0,
+                            "maxItems": 2,
+                            "items": {
+                                "type": "object",
+                                "required": [
+                                    "floor_name"
+                                ],
```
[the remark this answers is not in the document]
bpark1327: According to [HEScore Engineering Documentation](https://hes-documentation.labworks.org/home/api-definitions/api-methods/submit-inputs/zone_floor), `floor_name` looks to be the only required element for `zone_floor`. Other properties are conditionally required based on the validation rule, as shown below.
image
Am I missing something?

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2452 @@
...
+                                        "if": {
+                                            "properties": {
+                                                "floor_name": {"const": "floor1"}
+                                            },
+                                            "required": ["floor_name"]
+                                        },
+                                        "then": {
+                                            "required": ["floor_area"]
+                                        }
+                                    },
```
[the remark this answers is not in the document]
bpark1327: Please see the validation rule in the comment above (i.e., `floor_area` is required if `floor_name` == `floor1`). This is what I implemented here. But I think you're right. `floor_area` should be required for both.

hescorehpxml/schemas/hescore_json.schema.json
bpark1327: According to HEScore Engineering Documentation, `hvac_name` looks to be the only required element. Other conditional requirements are specified later. Please let me know if I am missing something.

lirainer:
@nmerket I'm concerned with how many places we enforce the same validation rules, which means multiple places to update and multiple opportunities for errors. Not counting the XSD, currently the inputs are validated by PNNL on submission and then again by our API as part of the calculate_base_building method. Adding this JSON schema adds a third step and then during creation of the JSON file yet another resulting in four places that need to be kept  in sync. There must be a better way?

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -0,0 +1,2754 @@
...
+                                "year": {
+                                    "$id": "#/properties/building/properties/systems/properties/domestic_hot_water/properties/year",
+                                    "type": "integer",
+                                    "minimum": 1972,
+                                    "description": "Year water heater was manufactured (only used if efficiency_method is shipment_weighted)"
+                                },
+                                "energy_factor": {
+                                    "$id": "#/properties/building/properties/systems/properties/domestic_hot_water/properties/energy_factor",
+                                    "type": "number",
+                                    "description": "Energy factor of water heater (only used if efficiency_method is user)"
```
shorowit, line 1569: This should say "only used if efficiency_method is user or uef"; there is an [outstanding API bug](https://github.com/NREL/hescore-analysis/issues/52#issuecomment-802273559) found by Yueyue that the documentation is incorrect.

nmerket (approved):
This looks good to me. Let's get it merged.

PR #160, opened 2021-07-27
Adding effective R-value lookup

### LINKED ISSUE #156: Handle Assembly R-values
Walls and Roof R-values in OpenStudio-HPXML are specified as assembly R-values, but hescore-hpxml expects layers with nominal R-values. So, most surfaces are coming through with R-0.

nmerket:
Fixes #156

## Pull Request Description

Some HPXML files have an `AssemblyRValue` element instead of `Insulation/Layer`. 

## Checklist

- [x] Update translation docs


nmerket:
Left to do:
- [ ] Roofs
- [ ] Ceilings
- [ ] Floors
- [ ] Knee Walls?
- [ ] Foundation Walls (or throw error on `AssemblyRValue` and fix if someone really needs it)
- [ ] Slabs (or throw error on `AssemblyRValue` and fix if someone really needs it)

nmerket:
Also address the "continuous- exterior" and "continuous - interior" installation types for rigid insulation.

One more wrinkle: in ResStock, for foundation walls we're not specifying insulation material. We may not be checking that for foundation walls, but we should see, but it may be overly restrictive for above grade walls.

nmerket:
@bpark1327 to add unit tests and request review.

hescorehpxml/hpxml2.py
```diff
@@ -94,14 +127,32 @@ def get_attic_area(self, attic, is_one_roof, footprint_area, v3_roofs, v3_b):
...
 
     def get_hescore_walls(self, b):
-        return self.xpath(b,
-                          'h:BuildingDetails/h:Enclosure/h:Walls/h:Wall[h:ExteriorAdjacentTo="ambient" or not(h:ExteriorAdjacentTo)]',  # noqa: E501
-                          aslist=True)
+        return self.xpath(
+            b, 'h:BuildingDetails/h:Enclosure/h:Walls/h:Wall\
+                [((h:ExteriorAdjacentTo="ambient" and not(contains(h:ExteriorAdjacentTo, "garage"))) or\
+                    not(h:ExteriorAdjacentTo)) and not(contains(h:InteriorAdjacentTo, "attic"))]',  # noqa: E501
+            aslist=True)
```
bpark1327, line 155: @nmerket This is what I changed so that the gable/garage walls gets ignored.

hescorehpxml/hpxml3.py
```diff
@@ -160,15 +213,41 @@ def get_attic_area(self, attic, is_one_roof, footprint_area, roofs, b):
...
-        return self.xpath(b,
-                          'h:BuildingDetails/h:Enclosure/h:Walls/h:Wall[h:ExteriorAdjacentTo="outside" or not('
-                          'h:ExteriorAdjacentTo)]',
-                          # noqa: E501
-                          aslist=True)
+        return self.xpath(
+            b, 'h:BuildingDetails/h:Enclosure/h:Walls/h:Wall\
+                [((h:ExteriorAdjacentTo="outside" and not(contains(h:ExteriorAdjacentTo, "garage"))) or\
+                    not(h:ExteriorAdjacentTo)) and not(contains(h:InteriorAdjacentTo, "attic"))]',  # noqa: E501
+            aslist=True)
```
bpark1327, line 250: @nmerket This is what I changed so that the v3 gable/garage walls gets ignored.

bpark1327:
@nmerket This PR is ready for your review. I cannot add you as a reviewer. I guess it is because you opened this PR.

hescorehpxml/base.py
```diff
@@ -1623,22 +1796,31 @@ def get_building_zone_wall(self, b, bldg_about):
...
+            wall_const_type_ext_finish_areas = defaultdict(float)
+            wallua = 0
+            walltotalarea = 0
+            for walld in hpxmlwalls[side]:
+                const_type = walld['assembly_code'][2:4]
+                ext_finish = walld['assembly_code'][6:8]
+                rvalue = int(walld['assembly_code'][4:6])
+                if walld['assembly_eff_rvalue'] is not None:
+                    eff_rvalue = walld['assembly_eff_rvalue']
+                else:
```
nmerket, line 1808: This whole section is quite messy from a long history of doing things in weird ways. I think we could greatly simplify this. The `wall_eff_rvalues` is this relic of a way to use center of cavity R-values to combine walls on a side of a house as [described here](https://hescore-hpxml.readthedocs.io/en/latest/translation/zone_wall.html#area-weighted-wall-r-value). We now have the actual assembly R-values (a more correct value to use to do this area-weighted calculation) for every wall construction. Let's get rid of the `wall_eff_rvalue` lookup. 

When you have nominal R-values in the walls and have determined an assembly code for each wall:
1. lookup the _assembly effective R-value_ for each from the csv file. 
2. Do an area weighted U-value calc to determine the overall assembly effective R-value for all the walls on that side of that house. 
3. Find the closest assembly code to that combined R-value (matching siding, etc)

When you have assembly R-values for each wall on a side of the house:
1. Use them directly to do an area weighted U-value calc and determine the combined assembly R-value. 
2. Lookup the closest assembly code to that combined R-value (matching siding, etc)

In fact, you could simplify this even further. Since you're passing the assembly code and an assembly R-value (if present) you could combine those two paths like so:
1. If the `assembly_eff_rvalue is None` look one up from the csv file (happens when you have only NominalRValues).
2. Do an area weighted U-value calc to determine the overall assembly effective R-value for all the walls on that side of that house. 
3. Find the closest assembly code to that combined R-value (matching siding, etc)

This would reduce this to like 10 lines. 💣

hescorehpxml/hpxml2.py
nmerket: Okay. Looks good. Not sure why it didn't work before, but I'm glad it works now.

hescorehpxml/hpxml3.py
```diff
@@ -109,6 +126,42 @@ def get_attic_floor_rvalue(self, attic, b):
...
+    def every_attic_floor_layer_has_nominal_rvalue(self, attic, b):
+        frame_floors = self.get_attic_floors(attic, b)
+        every_layer_has_nominal_rvalue = True  # Considered to have nominal R-value unless assembly R-value is used
+        for frame_floor in frame_floors:
+            frame_floor_layers = self.xpath(frame_floor, 'h:Insulation/h:Layer', aslist=True)
+            if frame_floor_layers:
+                for layer in frame_floor_layers:
+                    if self.xpath(layer, 'h:NominalRValue') is None:
+                        every_layer_has_nominal_rvalue = False
+                        break
```
nmerket, line 158:
```suggestion
            for layer in self.xpath(frame_floor, 'h:Insulation/h:Layer', aslist=True):
                if self.xpath(layer, 'h:NominalRValue') is None:
                    every_layer_has_nominal_rvalue = False
                    break
```

tests/test_translation.py
```diff
@@ -106,6 +106,9 @@ def test_house7(self):
     def test_house8(self):
         self._do_full_compare('house8')
 
+    def test_assembly_rvalue(self):
+        self._do_full_compare('hescore_min_assembly_rvalue')
+
```
nmerket, line 111: This is good. I didn't see a test that hits knee walls. In general, look at the coverage report and make sure all branches of your code are being hit.

nmerket (commented):
This is good. Thanks for doing this. See below for my comments. General comments are:

- Let's consistently give preference to the assembly R-values over nominal R-values if they're present. 
- I see where you are trying to use what we were doing before to combine walls. We should just jettison that approach. I proposed a simpler approach below that relies exclusively on assembly R-values.

examples/house8.json
```diff
@@ -27,7 +27,7 @@
           "roof_assembly_code": "rfwf00tg",
           "roof_color": "medium",
           "roof_type": "vented_attic",
-          "ceiling_assembly_code": "ecwf38",
+          "ceiling_assembly_code": "ecwf25",
```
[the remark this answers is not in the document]
bpark1327: This is because I changed the area weighted attic floor coc R-value calculation in #e848f7e. By this change, the weighted average R-value is calculated by weighting the U-values by area.

docs/source/translation/zone_wall.rst
```diff
@@ -60,6 +60,20 @@ inputs in HPXML to discrete inputs required for HEScore.
 
 .. _construction codes: https://docs.google.com/spreadsheet/pub?key=0Avk3IqpWXaRkdGR6cXFwdVJ4ZVdYX25keDVEX1pPYXc&output=html
 
+The wall R-value can be described by using nominal R-value or assembly R-value.
+If a user wish to use a nominal R-value, nominal R-value for all layers needs to be provided.
```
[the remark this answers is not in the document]
bpark1327: Ouch 🤦‍♂️

examples/resstock_hescore.xml
```diff
@@ -0,0 +1,653 @@
+<HPXML xmlns='http://hpxmlonline.com/2019/10' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance' schemaVersion='3.0'>
```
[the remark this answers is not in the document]
bpark1327: No, we didn't merge the resstock-hes branch. I copied one of the example files in the resstock-hes branch and modified it so that we can run it without resstock mode. I named the file `resstock_hescore.xml`. I agree that it is confusing. I should probably rename it to `assembly_rvalue.xml`.

hescorehpxml/base.py
```diff
@@ -1053,12 +1154,6 @@ def get_building_zone_roof(self, b, footprint_area):
             atticds.append(atticd)
 
             # Attic area
-            # Discussion need here: We are using is_one_roof to say, when there's no area specified in HPXML,
-            # we can use footprint area.Should we judge is_one_roof here by looking at element amounts of both
-            # Attic and Roof elements (to be 1)?
-            # I changed it to only look at attic element here because the current logic is more like Attic is
-            # on top of Roof, is it allowed to have roofs undefined in attic? I think those roofs not referred by
-            # attics should be ignored in our translation, right?
```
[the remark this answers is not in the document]
bpark1327: I didn't delete it. It was deleted in commit 3bb0d48.

docs/source/translation/zone_wall.rst
```diff
@@ -118,20 +132,13 @@ is selected.
 Finally, if neither of the above conditions are met, the wall is specified as
 simply "Wood Frame" in HEScore. 
 
-In all cases the R-value is summed for all insulation layers and the
-nearest discrete R-value from the list of possible R-values for that wall type
-is used. For walls with rigid foam sheathing, R-5 is subtracted from the
-nominal R-value sum to account for the R-value of the sheathing in the HEScore
-construction assembly.
-
```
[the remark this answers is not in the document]
bpark1327: It is happening if nominal R-value is used. I just moved it to lines 67-71. Please let me know if it doesn't make sense.

docs/source/translation/zone_wall.rst
```diff
@@ -118,20 +132,13 @@ is selected.
...
 Siding is selected according to the :ref:`siding map <sidingmap>`.
 
 Structural Brick
 ================
 
 If ``WallType/StructuralBrick`` is found in HPXML, one of the structural brick
-codes in HEScore is specified. The nearest R-value to the sum of all the
-insulation layer nominal R-values is selected.
+codes in HEScore is specified.
```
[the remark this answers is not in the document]
bpark1327: I moved it to the upper level at lines 67-71 as it applies to all wall types.

hescorehpxml/base.py
```diff
@@ -1163,36 +1266,59 @@ def get_building_zone_roof(self, b, footprint_area):
...
+                    knee_wall_coc_rvalue = knee_wall_rvalue + 1.8
+                    knee_wall_ua = knee_wall_area / knee_wall_coc_rvalue
+
+                    attic_floor_coc_rvalue = attic_floor_rvalue + 0.5
+                    attic_floor_adj_ua = knee_wall_ua + atticd['roof_area'] / attic_floor_coc_rvalue
+                    attic_floor_adj_area = knee_wall_area + atticd['roof_area']
+                    attic_floor_rvalue = (attic_floor_adj_area / attic_floor_adj_ua) - 0.5
+                    atticd['roof_area'] = attic_floor_adj_area
+                atticd['attic_floor_coc_rvalue'] = attic_floor_rvalue + 0.5
+            else:
```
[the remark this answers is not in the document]
bpark1327: @nmerket It brings me to a question... We are not erroring out if both nominal and assembly R-values are missing. Instead, we assume nominal R-0. Let me take a wall as an example. Please see [here](https://github.com/NREL/hescore-hpxml/blob/09694f404ab91130abdf0f0eabf598a116d7ddc9/hescorehpxml/base.py#L194-L203). `every_wall_layer_has_nominal_rvalue` will return True even if both nominal and assembly R-values. This is reflected in other surface types.
Do we want to error out if both R-values are missing?

hescorehpxml/base.py
bpark1327: I tried to address it in 0218ec0. It changed some wall codes in `house4.xml` and `house6.xml`. I assume this is expected as we change the approach.

hescorehpxml/base.py
```diff
@@ -1252,28 +1389,45 @@ def get_building_zone_roof(self, b, footprint_area):
...
-            roof_rvalue = min(list(roof_center_of_cavity_rvalues[atticd['roofconstype']][atticd['extfinish']].keys()),
-                              key=lambda x: abs(atticd['roof_coc_rvalue'] - roffset - x))
+            # Get Roof Code
+            if 'roof_assembly_rvalue' in atticd:
+                closest_roof_code, closest_code_rvalue = \
+                    min([(doe2code, code_rvalue)
+                        for doe2code, code_rvalue in self.roof_assembly_eff_rvalues.items()
+                        if doe2code[2:4] in attic_roofs_d['roofconstype'] and
+                        doe2code[6:8] == attic_roofs_d['extfinish']],
+                        key=lambda x: abs(x[1] - atticd['roof_assembly_rvalue']))
```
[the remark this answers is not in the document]
bpark1327: I wasn't able to think of a better way 😬

hescorehpxml/base.py
```diff
@@ -381,17 +442,27 @@ def get_attic_knee_wall_rvalue_and_area(self, attic, b):
...
+                        every_knee_wall_layer_has_nominal_rvalue = False
+                        break
+            assembly_eff_rvalue = convert_to_type(
+                float,
+                self.xpath(knee_wall, 'h:Insulation/h:AssemblyEffectiveRValue/text()', raise_err=False)
+            )
+            if assembly_eff_rvalue is not None:
+                rvalue = assembly_eff_rvalue
+            elif every_knee_wall_layer_has_nominal_rvalue:
+                rvalue = self.xpath(knee_wall, 'sum(h:Insulation/h:Layer/h:NominalRValue)')
```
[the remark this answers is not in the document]
bpark1327: They're assigned to the same variable name in this function, but they are handled differently depending on whether they are described by assembly R-values or nominal R-values (Please see [here](https://github.com/NREL/hescore-hpxml/blob/8d66a62c5c6bd3a5e3e0ec7da60fecc4e5f7bb80/hescorehpxml/base.py#L1301-L1322)).

bpark1327:
@nmerket Okay. I think it is ready for another review.

hescorehpxml/base.py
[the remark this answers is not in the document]
nmerket: Got it.

hescorehpxml/base.py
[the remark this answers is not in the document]
nmerket: We weren't before and I don't think we want to now. Let's keep the same logic where if both are missing we just assume R-0.

hescorehpxml/base.py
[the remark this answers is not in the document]
nmerket: It's all good. Let's just leave it as is.

examples/house9.xml
```diff
@@ -0,0 +1,785 @@
+<HPXML xmlns='http://hpxmlonline.com/2019/10' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance' schemaVersion='3.0'>
```
nmerket, line 1: What's this house9 for?

hescorehpxml/base.py
```diff
@@ -792,7 +849,7 @@ def get_footprint_area(bldg):
         for zone_floor in bldg['zone']['zone_floor']:
             if zone_floor['foundation_type'] == 'cond_basement':
                 cond_basement_floor_area += zone_floor['floor_area']
-        return int(python2round(old_div((floor_area - cond_basement_floor_area), stories)))
+        return math.floor((floor_area - cond_basement_floor_area) / stories)
```
nmerket, line 852: Oh that was a gross one. Thanks for fixing that. And all the other `old_div`s.

hescorehpxml/base.py
```diff
@@ -1010,7 +1067,53 @@ def get_building_about(self, b, p):
...
+    def get_assembly_eff_rvalues_dict(self, construction):
+        if construction in ['wall', 'roof', 'ceiling', 'floor']:
+            with open(os.path.join(thisdir, 'lookups', f'lu_{construction}_eff_rvalue.csv'), newline='') as f:
+                reader = csv.DictReader(f)
+                assembly_eff_rvalues = {}
+                for row in reader:
+                    assembly_eff_rvalues[row['doe2code']] = float(row['Eff-R-value'])
+            return assembly_eff_rvalues
+        else:
+            raise TranslationError(f'{construction}: Invalid building envelope construction.')
```
nmerket, line 1079: Looks good. 🏆

hescorehpxml/base.py
```diff
@@ -1010,7 +1067,53 @@ def get_building_about(self, b, p):
...
+
+    @property
+    def floor_assembly_eff_rvalues(self):
+        if self._floor_assembly_eff_rvalues is None:
+            self._floor_assembly_eff_rvalues = self.get_assembly_eff_rvalues_dict('floor')
+        return self._floor_assembly_eff_rvalues
+
     def get_building_zone_roof(self, b, footprint_area):
+
+        def get_predominant_roof_property(atticd, attic_key):
```
nmerket, line 1107: 👍

nmerket (commented):
This looks really good. A few notes:

The lookup folder needs to be added to [MANIFEST.in](https://packaging.python.org/guides/using-manifest-in/) so those files are available when the package is installed not in dev mode. **Update: addressed in ea28943**

I need to bring #170 up to date and merge it in here before we merge this. **Update: done.**

This has got me thinking since we simplified combining walls, could we do the same for floors above crawlspaces and attics and roof insulation? It feels like unfairly moving the goal posts, so maybe we catch this next time?

examples/house9.xml
bpark1327: House9 is for testing a complex house with 3 attics and the building enclosure is described by assembly R-values. I added it for convenience. I'm open to any suggestions.

bpark1327:
> This has got me thinking since we simplified combining walls, could we do the same for floors above crawlspaces and attics and roof insulation? It feels like unfairly moving the goal posts, so maybe we catch this next time?

I think we should do the same for floors and attics and roof insulation. And I guess now is the perfect time to address it. My question is, should we do it in this PR or should we open another PR?

nmerket:
Let’s do it in this PR. Get it all in at once.

Noel

________________________________

bpark1327:
@nmerket I updated the code for roof R-value. Could you please review it before I go further?

hescorehpxml/base.py
```diff
@@ -1255,13 +1246,21 @@ def get_predominant_roof_property(atticd, attic_key):
...
-                        roof_rvalue = max(roof_rvalue - 5, 0)
-                    roof_rvalue, attic_roofs_d['roof_coc_rvalue'] = \
-                        min(list(roof_center_of_cavity_rvalues[attic_roofs_d['roofconstype']][
-                                    attic_roofs_d['extfinish']].items()),
-                            key=lambda x: abs(x[0] - roof_rvalue))
+                    # For wood frame with radiant barrier roof, use wood frame roof effective R-value.
+                    # Radiant barrier will be handled by the actual radiant barrier model in OS.
+                    if attic_roofs_d['roofconstype'] == 'rb':
+                        roof_rvalue = 0
+                        roof_code = f"rfwf{roof_rvalue:02d}{attic_roofs_d['extfinish']}"
```
bpark1327, line 1253: For wood frame with radiant barrier roof, I suggest using wood frame roof effective R-value (without radiant barrier adjustment). In OS-HEScore, we changed the ruleset to [use the actual radiant barrier model in OS](https://github.com/NREL/OpenStudio-HEScore/pull/204). Using the adjusted R-value for the radiant barrier roof will double count the effect of the radiation barrier. Does it make sense?

nmerket: Does it make sense to still have assembly R-values available for these radiant barrier construction codes. Should we remove them from the table? @shorowit

docs/source/translation/zone_roof.rst
```diff
@@ -360,9 +322,17 @@ wall". See below an example:
...
-to the attic floor area. The knee walls center-of-cavity R-value is R-1.8 greater 
-than the nominal R-value. The knee walls center-of-cavity R-value will be reflected 
+to the attic floor area.
+
+The knee walls R-value can be described by nominal R-value or assembly R-value.
+
+If nominal R-value is used, the knee walls center-of-cavity R-value will be reflected 
 in the area weighted center-of-cavity effective R-value of the attic floor. 
-This averaged center-of-cavity effective R value is combined from all knee walls 
+The knee walls center-of-cavity R-value is R-1.8 greater than the nominal R-value.
```
nmerket, line 331: How are we doing the center-of-cavity nonsense with knee walls when we just got rid of it for roofs?

nmerket (commented):
A few comments below, but mostly looks good.

hescorehpxml/base.py
shorowit: I don't know what table you're referring to, but yes we should remove any references to the radiant barrier construction codes. I agree with @bpark1327's suggestion.

We should probably also relax the limitation of the roof being uninsulated (`roof_rvalue = 0` above) and remove that language from the documentation ("If the Roof/RadiantBarrier element exists and has a "true" value, the attic is assumed to have a radiant barrier **and no roof deck insulation is assumed according to the construction codes available in HEScore**.")

bpark1327: I'd like to make sure that we all are on the same page. I think the variable name, `roof_code`,  is confusing. It's more like a `lookup_code`. I use that code only for assembly R-value lookup. For a roof with a radiant barrier, I pass the code that has `rb` in it. This is because that seems to be the only way we can tell [OS-HEScore that the roof has a radiant barrier](https://github.com/NREL/OpenStudio-HEScore/blob/5e7016c2d6b851a1ea57d672282e5dca949511d6/rulesets/HEScoreRuleset/resources/HESruleset.rb#L184-L188). It's quite messy...

I guess the cleaner way to handle a roof with a radiant barrier would be to pass the code without radiant barrier (e.g., rf**wf**... instead of rf**rb**...) and add an element `has_radiant_barrier` boolean to the JSON file so OS-HEScore knows the roof has the radiant barrier when reading the JSON file. But I don't know if it is doable.

docs/source/translation/zone_roof.rst
bpark1327: I was going to handle knee walls when I update the code for attic floors because knee walls are reflected in attic floors. Here is what I was going to do:

_If knee walls are described by nominal R-value_:
1. Add knee wall's nominal R-value to the attic floor nominal R-value (get rid of coc adjustment for the knee wall)
2. Determine an assembly code for each attic floor
3. Lookup the assembly effective R-value for each from the csv file
4. Do an area-weighted U-value calc to determine the overall assembly effective R-value for all the attic floors
5. Find the closest assembly code to that combined R-value

_If knee walls are described by assembly R-value_:
1. Add knee wall's assembly R-value to the attic floor assembly R-value
2. Do an area-weighted U-value calc to determine the overall assembly effective R-value for all the attic floors
3. Find the closest assembly code to that combined R-value

Please let me know what you think. @nmerket

hescorehpxml/base.py
shorowit: Great points, @bpark1327, I forgot about some of the details here.

I suggest we discuss this with @lirainer during the call later today.

nmerket: Model as a roof without radiant barrier if R-value is > 0 and the radiant barrier is present in the HPXML. Only model with radiant barrier code if R-value = 0 and radiant barrier.

bpark1327:
It seems that the CI fails because of a new test file (i.e., `house9.xml`) that I added to this branch. `house9` is one of the example files generated by ResStock. `house9` includes 2 `DuctLeakageMeasurement` elements -- one for the supply and one for the return ducts. It appears that our workflow is not able to handle it properly. I don't know if it would make sense to calculate the area weighted duct leakage by `DuctSurfaceArea`. Please advise. @shorowit @nmerket

shorowit:
@bpark1327 I think we should support separate supply/return leakage values. Just need to add them together and then it's identical to the [existing `DuctLeakageMeasurement/DuctLeakage[TotalOrToOutside="to outside"]/Value` input](https://hescore-hpxml.readthedocs.io/en/latest/translation/hvac_distribution.html#duct-leakage-measurements).

docs/source/translation/zone_roof.rst
bpark1327: [Another PR](https://github.com/NREL/OpenStudio-HEScore/pull/320) will be handling the knee walls.

bpark1327:
> @bpark1327 I think we should support separate supply/return leakage values. Just need to add them together and then it's identical to the [existing `DuctLeakageMeasurement/DuctLeakage[TotalOrToOutside="to outside"]/Value` input](https://hescore-hpxml.readthedocs.io/en/latest/translation/hvac_distribution.html#duct-leakage-measurements).

Got it. I have a follow-up question. Do we always expect to get leakage values of CFM25? Or, do we need to support the values in other units (e.g., CFM50, Percent (CFM25/sqft), etc.)? @shorowit @nmerket

shorowit:
Just supporting CFM25 seems reasonable. We should definitely change `DuctLeakageMeasurement/DuctLeakage[TotalOrToOutside="to outside"]/Value` to `DuctLeakageMeasurement/DuctLeakage[TotalOrToOutside="to outside" and Units="CFM25"]/Value`, for example, so that we don't misuse the provided value.

hescorehpxml/base.py
```diff
@@ -625,8 +625,8 @@ def get_hvac_distribution(self, hvacd_el, bldg):
         # Duct leakage to outside
         leakage_to_outside = \
             self.xpath(airdist_el,
-                       'h:DuctLeakageMeasurement/h:DuctLeakage[h:TotalOrToOutside="to outside" and Units="CFM25"]' +
-                       '/h:Value/text()')
+                       'sum(h:DuctLeakageMeasurement/h:DuctLeakage[h:TotalOrToOutside="to outside" ' +
+                       'and h:Units="CFM25"]/h:Value/text())')
 
         if leakage_to_outside:
```
shorowit, line 631: I don't think this is very robust:
1. If the HPXML file has, say, 3 DuctLeakageMeasurement elements (supply, return, and the sum of them), it would end up with twice the desired value. I think we should explicitly look for 1 DuctLeakageMeasurement  where DuctType is not provided or 2 DuctLeakageMeasurements where DuctType is supply and return.
2. I believe that switching to sum() in the xpath will now return a value of zero if there are no DuctLeakageMeasurement elements. This likely doesn't break any tests because the line below is `if leakage_to_outside:`, which evaluates to false if `leakage_to_outside` is zero (just as it previously evaluated to false if `leakage_to_outside` was `None`). But that means that someone can't enter a duct leakage to outside of zero because it won't get treated as qualitative leakage. In short, the code really needs to distinguish between no DuctLeakageMeasurement elements (qualitative), DuctLeakageMeasurement elements that sum to zero (quantitative). And `if leakage_to_outside:` should be changed to `if leakage_to_outside is not None:`. It's a bad idea to use the `if foo:` construct unless the variable being evaluated is boolean and only boolean.

Also don't forget to update the documentation.

bpark1327: These are really good points. Especially, the second comment is something that didn't even think about. I'll keep it in mind.

nmerket: Yes, very good points, @shorowit. Thanks for the thorough review.

bpark1327:
@shorowit Duct blaster measurement will be updated in #183

bpark1327:
@nmerket I think it's ready for your review. Please let me know if you have any questions.

hescorehpxml/hpxml2.py
```diff
@@ -81,6 +98,22 @@ def get_attic_type(self, attic, atticid):
     def get_attic_floor_rvalue(self, attic, v3_b):
         return self.xpath(attic, 'sum(h:AtticFloorInsulation/h:Layer/h:NominalRValue)')
 
+    def get_attic_floor_assembly_rvalue(self, attic, v3_b):
+        return self.xpath(attic, 'h:AtticFloorInsulation/h:AssemblyEffectiveRValue/text()')
```
nmerket, line 102: I saw that you were wrapping this call with the `convert_to_float` function. I think it makes more sense to put that _inside_ this one so that it returns a number. Obviously you'd need to do the same in hpxml3.py.

tests/test_translation.py
```diff
@@ -867,8 +870,8 @@ def test_attic_knee_wall(self):
         attic_type.addprevious(etree.Element(tr.addns('h:AtticKneeWall'), {'idref': 'wall2'}))
         # run translation
         resp = tr.hpxml_to_hescore()
-        self.assertEqual(resp['building']['zone']['zone_roof'][0]['ceiling_assembly_code'], 'ecwf38')
-        self.assertAlmostEqual(resp['building']['zone']['zone_roof'][0]['roof_area'], 1400.0)
+        self.assertEqual(resp['building']['zone']['zone_roof'][0]['ceiling_assembly_code'], 'ecwf49')
+        self.assertAlmostEqual(resp['building']['zone']['zone_roof'][0]['roof_area'], 1200.0)
```
nmerket, line 874: So, this is removing the knee wall area and R-value from the test expectation? That seems like what we want.

nmerket (commented):
I think this is looking really good. Just a few notes.

examples/house8.json
```diff
@@ -27,7 +27,7 @@
           "roof_assembly_code": "rfwf00tg",
           "roof_color": "medium",
           "roof_type": "vented_attic",
-          "ceiling_assembly_code": "ecwf38",
+          "ceiling_assembly_code": "ecwf21",
```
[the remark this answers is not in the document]
bpark1327: This is because I fixed area-weighted R-value calculation of attic floors in #e848f7e. Previously, the area-weighted R-value was not calculated by weighting the U-values by area.

hescorehpxml/base.py
```diff
@@ -567,7 +625,8 @@ def get_hvac_distribution(self, hvacd_el, bldg):
         # Duct leakage to outside
         leakage_to_outside = \
             self.xpath(airdist_el,
-                       'h:DuctLeakageMeasurement/h:DuctLeakage[h:TotalOrToOutside="to outside"]/h:Value/text()')
+                       'sum(h:DuctLeakageMeasurement/h:DuctLeakage[h:TotalOrToOutside="to outside" ' +
+                       'and h:Units="CFM25"]/h:Value/text())')
```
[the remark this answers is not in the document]
bpark1327: Sorry for the confusion. I had to revert it. This code will be replaced with the changes I made in #183.

tests/test_translation.py
bpark1327: Correct. That's why we see greater R-values and smaller roof areas.

hescorehpxml/base.py
```diff
@@ -1037,29 +1096,64 @@ def get_building_about(self, b, p):
...
+    def get_assembly_eff_rvalues_dict(self, construction):
+        if construction in ['wall', 'roof', 'ceiling', 'floor']:
+            with open(os.path.join(thisdir, 'lookups', f'lu_{construction}_eff_rvalue.csv'), newline='') as f:
+                reader = csv.DictReader(f)
+                assembly_eff_rvalues = {}
+                for row in reader:
+                    assembly_eff_rvalues[row['doe2code']] = float(row['Eff-R-value'])
+            return assembly_eff_rvalues
+        else:
+            raise TranslationError(f'{construction}: Invalid building envelope construction.')
```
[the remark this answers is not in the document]
bpark1327: I think you're right. We'd probably never hit this code in normal operation. Let's use `assert` here.

bpark1327:
@nmerket I addressed your comments. I moved some functions for walls to hpxml2 and hpxml3.py to be consistent. Please let me know what you think.

nmerket (commented):
Looks good to me. I think I've put you through enough on this. Let's merge it. I can't approve this because I originally opened the PR. Maybe @yzhou601 could give it a once over?

PR #161, opened 2021-08-03
JSON Schema, take 2

bpark1327:
## Pull Request Description

- bugfixes
- add additional properties
    - [ ] efficiency level for heating, cooling, and domestic hot water systems

## Checklist

- [ ] Update translation docs

### DIFF
--- a/hescorehpxml/base.py
+++ b/hescorehpxml/base.py
@@ -1657,7 +1657,7 @@ class HPXMLtoHEScoreTranslatorBase(object):
             windowd = {'area': convert_to_type(float, xpath(hpxmlwndw, 'h:Area/text()', raise_err=True))}
             windowd['uvalue'] = convert_to_type(float, xpath(hpxmlwndw, 'h:UFactor/text()'))
             windowd['shgc'] = convert_to_type(float, xpath(hpxmlwndw, 'h:SHGC/text()'))
-            windowd['sun_screen'] = self.get_sunscreen(hpxmlwndw)
+            windowd['solar_screen'] = self.get_sunscreen(hpxmlwndw)
             if windowd['uvalue'] is not None and windowd['shgc'] is not None:
                 windowd['window_code'] = None
             else:
@@ -1771,7 +1771,7 @@ class HPXMLtoHEScoreTranslatorBase(object):
                 zone_window['window_area'] = 0
                 zone_window['window_method'] = 'code'
                 zone_window['window_code'] = 'scna'
-                zone_window['sun_screen'] = False
+                zone_window['solar_screen'] = False
                 continue
 
             # Get the list of uvalues and shgcs for the windows on this side of the house.
@@ -1815,9 +1815,9 @@ class HPXMLtoHEScoreTranslatorBase(object):
             window_sunscreen_areas = {}
             for window in windows:
                 try:
-                    window_sunscreen_areas[window['sun_screen']] += window['area']
+                    window_sunscreen_areas[window['solar_screen']] += window['area']
                 except KeyError:
-                    window_sunscreen_areas[window['sun_screen']] = window['area']
+                    window_sunscreen_areas[window['solar_screen']] = window['area']
             zone_window['solar_screen'] = max(list(window_sunscreen_areas.items()), key=lambda x: x[1])[0]
         return zone_wall
 
--- a/hescorehpxml/schemas/hescore_json.schema.json
+++ b/hescorehpxml/schemas/hescore_json.schema.json
@@ -2,6 +2,7 @@
     "$schema": "http://json-schema.org/draft-07/schema",
     "type": "object",
     "title": "HEScore JSON Schema",
+    "additionalProperties": false,
     "properties": {
         "building_address": {
             "$id": "#/properties/building_address",
@@ -39,6 +40,7 @@
         "hpwes": {
             "$id": "#/properties/hpwes",
             "type": "object",
+            "additionalProperties": false,
             "properties": {
                 "improvement_installation_start_date": {
                     "$id": "#/properties/building/properties/improvement_installation_start_date",
@@ -73,11 +75,11 @@
         "building": {
             "$id": "#/properties/building",
             "type": "object",
+            "additionalProperties": false,
             "properties": {
                 "about": {
                     "$id": "#/properties/building/properties/about",
                     "type": "object",
-                    "additionalProperties": false,
                     "required": [
                         "assessment_date",
                         "shape",
@@ -89,6 +91,7 @@
                         "orientation",
                         "blower_door_test"
                     ],
+                    "additionalProperties": false,
                     "properties": {
                         "assessment_date": {
                             "$id": "#/properties/building/properties/about/properties/assessment_date",
@@ -240,6 +243,7 @@
                         "wall_construction_same",
                         "window_construction_same"
                     ],
+                    "additionalProperties": false,
                     "properties": {
                         "zone_roof": {
                             "$id": "#/properties/building/properties/zone/properties/zone_roof",
@@ -651,6 +655,7 @@
                                     "zone_window": {
                                         "type": "object",
                                         "description": "Inputs about the windows on this wall",
+                                        "additionalProperties": false,
                                         "properties": {
                                             "window_area": {
                                                 "type": "number",
@@ -710,6 +715,7 @@
                                 "required": [
                                     "hvac_name"
                                 ],
+                                "additionalProperties": false,
                                 "properties": {
                                     "hvac_name": {
                                         "$id": "#/properties/building/properties/systems/properties/hvac/items/properties/hvac_name",
@@ -733,6 +739,7 @@
                                         "$id": "#/properties/building/properties/systems/properties/hvac/items/properties/heating",
                                         "type": "object",
                                         "description": "Inputs about the heating system",
+                                        "additionalProperties": false,
                                         "properties": {
                                             "type": {
                                                 "$id": "#/properties/building/properties/systems/properties/hvac/items/properties/heating/properties/type",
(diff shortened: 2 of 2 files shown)


hescorehpxml/base.py
```diff
@@ -1771,7 +1771,7 @@ def windows_are_on_shared_walls():
                 zone_window['window_area'] = 0
                 zone_window['window_method'] = 'code'
                 zone_window['window_code'] = 'scna'
-                zone_window['sun_screen'] = False
+                zone_window['solar_screen'] = False
```
bpark1327, line 1774: @nmerket This was causing JSON schema validation failure. I changed `sun_screen` to `solar_screen` in `base.py`. Please let me know if what I have done is wrong.

shorowit: @lirainer Is the API looking for `sun_screen` or `solar_screen` in the JSON files from hescore-hpxml?

lirainer: Looks like it uses the window elements as submitted:
```
            $inputs->zone->zone_wall[$id]->zone_window = (object) $inputs->zone->zone_wall[$id]->zone_window;
```
Which means is will expect `solar_screen`. But I will run a test.

lirainer: The version of hescore-hpxml that I am running (6.1.0 ) returns `solar_screen` as it should:
```
      "zone_wall": [
        {
          "side": "front", 
          "wall_assembly_code": "ewwf13vi", 
          "zone_window": {
            "window_area": 60.0, 
            "window_method": "code", 
            "window_code": "dcaw", 
            "solar_screen": true
          }
        }, 
```
And that gets read correctly by our code.

lirainer: How did `sun_screen` sneak in there?

shorowit: @yzhou601 do you have any idea what is going on here?

lirainer:
@shorowit @bpark1327 Here is a set of JSON regression files that all pass the current schema: 
[v9_0_210808.zip](https://github.com/NREL/hescore-hpxml/files/6955357/v9_0_210808.zip)

bpark1327:
> 
> 
> @shorowit @bpark1327 Here is a set of JSON regression files that all pass the current schema:
> [v9_0_210808.zip](https://github.com/NREL/hescore-hpxml/files/6955357/v9_0_210808.zip)

@lirainer Thanks for sending these files over. I was able to find some bugs in the code by running them. I have just noticed that all files have `hvac_fraction` of 0. Could you please confirm?

lirainer:
@bpark1327 Yup, that is an error on our part

bpark1327:
@lirainer Here are additional issues I found:
- In `Duct_multiple.json`, there are 3 ducts and all of them have `fraction` of 1 (sum of duct fractions = 3). As far as I know, this kind of checking cannot be done in the standard JSON schema unfortunately. @nmerket please correct me if I am wrong.
- In `HVAC2.json`, it seems that the `hvac_fraction` of systems has changed. I think `hvac_fraction` of `hvac1` was 0.7 and that of `hvac2` was 0.3 previously.

hescorehpxml/base.py
nmerket: This is a mistake. Just use `solar_screen`.

lirainer:
@bpark1327 Here is a new set with fixes for the issues you identified: [v9_0_210816.zip](https://github.com/NREL/hescore-hpxml/files/6993950/v9_0_210816.zip)

shorowit:
@lirainer The HES API uses 0-1 for HVAC fractions and 0-100 for HVAC distribution fractions, which we're now further perpetuating in the JSON schema. Is there an opportunity now to address this in the API and JSON schema? Or is it too painful to change at this point and I should pretend I never saw this inconsistency?

For what it's worth, I imagine it _might_ be possible to make a non-breaking API change and accommodate both. For example, change from integer to float but still allow 0-100. The JSON schema and hescore-hpxml would use 0-1. If the API receives a value <= 1, it would multiply by 100, otherwise it uses the value directly.

lirainer:
@shorowit I would say the pain comes from kicking myself every time I see this inconsistency. The `hvac_distribution/fraction` is an int because we used an existing integer `length` field in our hvac_distribution table to store the fraction rather than add a new field. At a minimum we should make the JSON schema consistently 0-1. Changing the field type in our table is not an issue, but we should discuss whether it is worth making a breaking change in the API. Not sure if I like your suggestion of overloading the type - seems that will lead to problems.

bpark1327:
@lirainer here are a few more issues I found:
- There seems to be wall(s) that should not be in the townhouse (e.g., a "left wall" in Townhouse_left.json)
- `Duct_uninsulated_and_unsealed.json` has an insulated duct
- `Window_area_0.json` has a window on the front wall

lirainer:
@bpark1327 Thanks. We will fix the first two. But not the third one:
> * `Window_area_0.json` has a window on the front wall
This intensionally has the front window area set to 1 as an HPXML file with no windows does not pass validation.

bpark1327:
> This intensionally has the front window area set to 1 as an HPXML file with no windows does not pass validation.

Ah I see. Thanks for the information. I will change the HESruleset to handle it.

shorowit:
@lirainer When you send the updated JSON files, could you also include [ENERGY STAR ones](https://github.com/NREL/OpenStudio-HEScore/issues/264)? (Or should we just manually create them ourselves?)

lirainer:
@shorowit Yes, I think that would be easy to do. I will add them on or let you know if there are any issues.

shorowit:
Thanks @lirainer. We're happy to wait a week for the files from you.

lirainer:
@shorowit Here are the ES JSON runs: 
[energystar.zip](https://github.com/NREL/hescore-hpxml/files/7040132/energystar.zip)
Running them turned out to be easy. But they did turn up an error in the code, so glad you remembered about them.

bpark1327:
> 
> 
> @shorowit Here are the ES JSON runs:
> [energystar.zip](https://github.com/NREL/hescore-hpxml/files/7040132/energystar.zip)
> Running them turned out to be easy. But they did turn up an error in the code, so glad you remembered about them.

@lirainer `ES_boiler_fuel_oil` has a heating system, as shown below:
```
"heating": {
    "fuel_primary": "fuel_oil",
    "type": "boiler",
    "efficiency_method": "energy_star"
},
```
In my understanding, what we expect to see is something like:
```
"heating": {
    "fuel_primary": "fuel_oil",
    "type": "boiler",
    "efficiency_level": "energy_star"
},
```
I haven't seen all ES files but they probably have the similar issue. Please let me know if I am missing something.

lirainer:
@bpark1327 Fixed `efficiency_level`. Also found DHW was not being done correctly. Here is a new set: 
[energystar.zip](https://github.com/NREL/hescore-hpxml/files/7041680/energystar.zip)
None of these pass validation with the schema, but I think they are now correct.

bpark1327:
> None of these pass validation with the schema, but I think they are now correct.

@lirainer I fixed the schema and ES JSON files should pass validation.

bpark1327:
@lirainer I realized that there is a distribution system for the mini-split in `ES_mini_split_electric.json`. Whereas, there is no distribution system for the mini-split in `HVAC_mini_split.json`, which makes more sense to me. Is the distribution system in `ES_mini_split_electric.json` intended?

nmerket:
Add [electric wall furnace](https://github.com/NREL/OpenStudio-HEScore/issues/292#issuecomment-911802976).

bpark1327:
[Per Leo](https://github.com/NREL/OpenStudio-HEScore/pull/261#discussion_r710578623), the JSON schema should allow for both the blower door value and the sealed value at the same time.

examples/hescore_min.json
```diff
@@ -111,7 +111,7 @@
             {
               "name": "duct1",
               "location": "cond_space",
-              "fraction": 100,
+              "fraction": 1.0,
```
nmerket, line 114: Using floats now ✅

hescorehpxml/base.py
```diff
@@ -603,14 +603,14 @@ def get_hvac_distribution(self, hvacd_el, bldg):
             hvacd_out = OrderedDict()
             hvacd_out['name'] = 'duct%d' % i
             hvacd_out['location'] = hvacd['location']
-            hvacd_out['fraction'] = int(python2round(hvacd['fraction'] / sum_of_top_3_fractions * 100))
+            hvacd_out['fraction'] = hvacd['fraction'] / sum_of_top_3_fractions
```
nmerket, line 606: Thanks for cleaning this up and getting rid of the artifacts from the 2to3 conversion.

hescorehpxml/base.py
```diff
@@ -2054,7 +2054,7 @@ def iter_next(_iter):
             return retval
 
         def choose_dist_system(first_choice_dist_id, second_choice_dist_id):
-            if first_choice_dist_id is not None and len(distribution_systems[first_choice_dist_id]) > 0:
+            if first_choice_dist_id is not None and distribution_systems[first_choice_dist_id] is not None:
```
nmerket, line 2057: Is there ever a case anymore where we'd have this be an empty list? If so, does this handle that correctly?

bpark1327: I'd say I am 99% sure that we'd not have it be an empty list. I changed [get_hvac_distribution method](https://github.com/NREL/hescore-hpxml/blob/975aca5f1c0cd34da97a3c911699b78e5f6d170d/hescorehpxml/base.py#L554-L555) and removed the code in lines 2128-2129. 
If we are concerned about the 1%, we can add back len() > 0. Please let me know what you think.

bpark1327:
@nmerket As I noticed [here](https://github.com/NREL/hescore-hpxml/pull/166#issuecomment-932724708), the latest `jsonschema` package for Python 3.7+ seems to generate different error messages. It caused CI failure. Should I force it to use an older version?

hescorehpxml/base.py
nmerket: This is probably fine as is. I just wanted to check with you about it to make sure you'd thought through it.

nmerket:
@bpark1327 It would be better to find out the underlying reason_why_ the new version breaks and fix it. At some point we'll need to update. Might as well do it now.

bpark1327:
> 
> 
> @bpark1327 It would be better to find out the underlying reason_why_ the new version breaks and fix it. At some point we'll need to update. Might as well do it now.

As we discussed, we will use `jsonschema`  version 3.1.1 (older version) to avoid CI test failures.

nmerket (approved):

PR #163, opened 2021-09-08
Update duct insulation mapping

### LINKED ISSUE #141: Verify that hpxml2hescore works on workflow HPXML
The hpxml2hescore script should be able to process workflow HPXML without loosing any information. Run all the regression set HPXML files through scoring tool API submit_hpxml() method and compare the results to the regression set results and Identify any differences.

I have identified one issue so far:

HEScore designates insulated ducts using  `DuctInsulationMaterial/Unknown` and uninsulated ducts using `DuctInsulationMaterial/None`. But hpxml2hescore maps these both to `"insulated": false.`

bpark1327:
Fixes #141 

## Pull Request Description

Map ducts with `DuctInsulationMaterial` that is not `None` to insulated ducts

## Checklist

- [x] Update translation docs

### DIFF
--- a/docs/source/translation/hvac_distribution.rst
+++ b/docs/source/translation/hvac_distribution.rst
@@ -81,8 +81,8 @@ Duct Insulation
 
 If the any of the ``Ducts`` elements in a particular
 :ref:`location <ductlocationmapping>` have a ``DuctInsulationRValue`` or
-``DuctInsulationThickness`` that is greater than zero, all of the ducts in that
-location are considered insulated.
+``DuctInsulationThickness`` that is greater than zero or have a ``DuctInsulationMaterial/Unknown``, 
+all of the ducts in that location are considered insulated.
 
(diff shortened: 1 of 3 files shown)


nmerket (commented):
Could you add a test that hits this?

shorowit:
If you're going to map 'Unknown' to insulated ducts, wouldn't it make more sense to map all values (e.g., 'Batt', 'LooseFill') other than 'None' to insulated ducts?

nmerket:
Good catch, Scott. Yes, any insulation material besides "None" should be mapped to insulated ducts

lirainer:
@bpark1327 I was out last week so I'm catching up. Seems to me that this will break if a new insulation material is added to HPXML. Wouldn't it be better to use something like `... or DuctInsulationMaterial exists and != None`?

nmerket (commented):
Looks good. One change below.

PR #179, opened 2021-12-09
PV Number of Panels

### LINKED ISSUE #178: Read NumberOfPanels from PV in HPXML v3
**Is your feature request related to a problem? Please describe.**
The `NumberOfPanels` element was added in HPXML v3 (probably to support HEScore). Currently we're either reading from `MaxPowerOutput` or estimating the number of panels from `CollectorArea` because that input wasn't available in HPXML v2. Now that we have it, we should read it. 

**Describe the solution you'd like**
The following logic:

- If `MaxPowerOutput` is available, use that.
- Else if `NumberOfPanels` is available, use that.
- Else if `CollectorArea` is available, estimate number of panels from that as we do now.

cc @torstenglidden

nmerket:
Fixes #178

## Pull Request Description

Allows `NumberOfPanels` to be used directly from an HPXML (v3) file.

## Checklist

- [x] Update translation docs

### DIFF
--- a/hescorehpxml/base.py
+++ b/hescorehpxml/base.py
@@ -65,6 +65,10 @@ def round_to_nearest(x, vals, tails_tolerance=None):
     return nearest
 
 
+def weighted_average(items, weights):
+    return sum(item * weight for item, weight in zip(items, weights)) / sum(weights)
+
+
 class HPXMLtoHEScoreTranslatorBase(object):
     SCHEMA_DIR = None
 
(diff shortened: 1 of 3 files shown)


hescorehpxml/base.py
```diff
@@ -2235,31 +2239,28 @@ def get_generation(self, b):
...
-                else:
-                    raise TranslationError('MaxPowerOutput or CollectorArea is required for every PVSystem.')
+            capacities.append(convert_to_type(float, self.xpath(pvsystem, 'h:MaxPowerOutput/text()')))
+            collector_areas.append(convert_to_type(float, self.xpath(pvsystem, 'h:CollectorArea/text()')))
+            n_panels_per_system.append(convert_to_type(int, self.xpath(pvsystem, 'h:NumberOfPanels/text()')))
+
+            if not (capacities[-1] or collector_areas[-1] or n_panels_per_system[-1]):
+                raise TranslationError(
+                    'MaxPowerOutput, NumberOfPanels, or CollectorArea is required for every PVSystem.'
+                )
```
bpark1327, line 2255: It'd be an edge case, but this conditional statement seems to have room to refer to the elements that are populated by other PV systems.

bpark1327 (commented):
@nmerket I have a question for you. Please see below.

hescorehpxml/base.py
nmerket: I don't think so because on lines 2248-2250 I'm adding a value to the end of each list so the last value on the list is always the one for the current pv system.

bpark1327: Oh, I see. It will append `None` if missing. So the last values will always be the ones for the current PV system.

bpark1327 (approved):
Looks good to me! 👍

PR #185, opened 2022-01-06
CEE HVAC efficiency levels

yzhou601:
Fixes #

## Pull Request Description

description here

## Checklist

- [ ] Update translation docs

### DIFF
--- a/hescorehpxml/schemas/hescore_json.schema.json
+++ b/hescorehpxml/schemas/hescore_json.schema.json
@@ -1028,7 +1028,7 @@
                                                         "properties": {
                                                             "heating": {
                                                                 "properties": {
-                                                                    "efficiency_level": {"const": "energy_star"}
+                                                                    "efficiency_level": {"enum": ["energy_star", "cee_tier1", "cee_tier2", "cee_tier3"]}
                                                                 },
                                                                 "required": ["efficiency_level"]
                                                             }
@@ -1243,7 +1243,7 @@
                                                         "properties": {
                                                             "cooling": {
                                                                 "properties": {
-                                                                    "efficiency_level": {"const": "energy_star"}
+                                                                    "efficiency_level": {"enum": ["energy_star", "cee_tier1", "cee_tier2", "cee_tier3"]}
                                                                 },
                                                                 "required": ["efficiency_level"]
                                                             }
--- a/hescorehpxml/schemas/hescore_json.schema.json
+++ b/hescorehpxml/schemas/hescore_json.schema.json
@@ -708,7 +708,7 @@
                                             },
                                             "efficiency_level": {
                                                 "type": "string",
-                                                "enum": ["energy_star"],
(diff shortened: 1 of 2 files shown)


shorowit:
@bpark1327 will help with review and development of unit tests.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -829,6 +849,23 @@
                                                             "maximum": 40,
                                                             "description": "SEER of cooling equipment (only used if efficiency_method is user)"
                                                         }
+                                                    },
+                                                    "if":{
+                                                        "properties": {"type": {"const": "heat_pump"}}
```
bpark1327, line 854: You should add `"required": ["type"]` to make this IF condition works properly. Please do the same thing for `mini-split`.

bpark1327:
As for the unit tests, you will find a bunch of tests in `test_schema.py`. I think you could add your tests in `test_invalid_heating` or `test_invalid_cooling`. Those tests deep copy two example files and make them invalid by changing or deleting some properties, then assert you get the error messages you expect to see. Please ping me if you have any questions.

hescorehpxml/schemas/hescore_json.schema.json
yzhou601: This is actually got me a bit confused. Does `"required": ["type"]` mean the type element is a required one? If so, isn't the outer if already taking care of that?

bpark1327: According to the JSON schema reference, the “required” keyword is necessary in the “if” schemas or they would all apply if the property (in our case `type`) is not defined. If you don't want this error to be triggered when `type` is not defined (note that `type` is not a required property for `heating` and `cooling`), you will need to add the "required" keyword.

bpark1327: FYI, https://json-schema.org/understanding-json-schema/reference/conditionals.html

yzhou601: No I mean in the outer "if" statement the `type` is already defined as required. Like [this](https://github.com/NREL/hescore-hpxml/blob/master/hescorehpxml/schemas/hescore_json.schema.json#L821-L823), so my logic is implemented inside of the previous "if", inside `if type is "split_dx" or "heat_pump"`, then `if type is "split_dx"`, is it necessary to have the second "if" also specifying the required `type`?

bpark1327: Oh, I see. I just realized that your "if" statement is within the "then" statement and you're using the `type` again in the "if" statement. I think you should be fine then. Because the "then" statement that is wrapping your "if" statement will not be triggered when the system type is neither "heat_pump" nor "mini-split". You can disregard this comment.

yzhou601:
@bpark1327 Could you review this PR, especially the schema changes that I made? @nmerket Could you review this PR too?

bpark1327 (approved):
Looks good to me! 👍

PR #198, opened 2022-04-21
Release v7.0

### LINKED ISSUE #192: Changelog
Write up a changelog for this latest version of the translator.

nmerket:
Fixes #192

## Pull Request Description

This updates the version to 7.0 in conjunction with the OpenStudio-HEScore v2.0 release. 

Also includes a changelog describing the updates. 

## Checklist

- [x] ~~Code changes (must work)~~
- [x] ~~Test exercising your feature or bug fix. Check the coverage report in the build artifacts.~~
- [x] Update translation docs


nmerket:
@torstenglidden Can you give this changelog a review? You can download the full docs [here](https://github.com/NREL/hescore-hpxml/suites/6212329655/artifacts/219590898).

torstenglidden (commented):
@nmerket The changelog generally looks good to me, but with the following comments/questions:

1. There were two other changes we announced that I didn't see changelog entries for:
     a) Low-e storm windows
     b) Added HVAC enumeration -- high-efficiency/cold climate heat pumps

2. In the HVAC section, in the linked description for Duct Leakage Measurements, the 2nd bulleted item in the seemed confusing to me...it seems to cut of short of saying what happens -- e.g., 

    If the above element is not found, DuctLeakageMeasurement[DuctType="supply"]/DuctLeakage[TotalOrToOutside="to outside" and Units="CFM25"]/Value and DuctLeakageMeasurement[DuctType="return"]/DuctLeakage[TotalOrToOutside="to outside" and Units="CFM25"]/Value elements with the numeric values (Note that the sum of two elements will be used)

-- I'm guessing it should read something like "If the above element is not found, then  DuctLeakageMeasurement[DuctType="supply"]/DuctLeakage[TotalOrToOutside="to outside" and Units="CFM25"]/Value   and   DuctLeakageMeasurement[DuctType="return"]/DuctLeakage[TotalOrToOutside="to outside" and Units="CFM25"]/Value elements with the numeric values will be used (Note that the sum of the two elements will be used)". Would that still be an accurate description (it seems clearer to me that way) or am I missing something else?

3. In the Water Heating section, I think it would be helpful to add "gas" before both instances of "water heater", just to avoid confusion with regard to the higher EF for heat pump water heaters and the "maximum allowed".

nmerket:
> @nmerket The changelog generally looks good to me, but with the following comments/questions:
> 
> 1. There were two other changes we announced that I didn't see changelog entries for:
>    a) Low-e storm windows
>    b) Added HVAC enumeration -- high-efficiency/cold climate heat pumps

Both of these were added as upgrades only, so they don't really affect the behavior this translator code. A user can specify a high efficiency heat pump already in the base building by indicating an appropriate HSPF and SEER. Same for Low-e storms using U-factor and SHGC, although I think we should discuss in our meeting today whether we should try to translate a low-e storm more for a base building from some elements in HPXML. 

> 2. In the HVAC section, in the linked description for Duct Leakage Measurements, the 2nd bulleted item in the seemed confusing to me...it seems to cut of short of saying what happens -- e.g.,
>    If the above element is not found, DuctLeakageMeasurement[DuctType="supply"]/DuctLeakage[TotalOrToOutside="to outside" and Units="CFM25"]/Value and DuctLeakageMeasurement[DuctType="return"]/DuctLeakage[TotalOrToOutside="to outside" and Units="CFM25"]/Value elements with the numeric values (Note that the sum of two elements will be used)
> 
> -- I'm guessing it should read something like "If the above element is not found, then DuctLeakageMeasurement[DuctType="supply"]/DuctLeakage[TotalOrToOutside="to outside" and Units="CFM25"]/Value and DuctLeakageMeasurement[DuctType="return"]/DuctLeakage[TotalOrToOutside="to outside" and Units="CFM25"]/Value elements with the numeric values will be used (Note that the sum of the two elements will be used)". Would that still be an accurate description (it seems clearer to me that way) or am I missing something else?

Good catch. I'll fix that.

> 
> 3. In the Water Heating section, I think it would be helpful to add "gas" before both instances of "water heater", just to avoid confusion with regard to the higher EF for heat pump water heaters and the "maximum allowed".

This also applies to electric resistance storage and tankless water heaters as well, but you're correct that it should be clarified that it does not apply to heat pump water heaters. I'll make that change.

PR #216, opened 2023-02-22
Geometry and Envelope Additions

bpark1327:
Fixes #

## Pull Request Description

Reflect changes applied to [a PR in OS-HEScore](https://github.com/NREL/OpenStudio-HEScore/pull/367)

## Checklist

- [x] Update translation docs


bpark1327:
@nmerket I don't know why test (3.6) fails. Please advise.

nmerket:
Python 3.6 is [end of life](https://devguide.python.org/versions/) and no longer getting support. Let's remove it from the [python version list](https://github.com/NREL/hescore-hpxml/blob/geometry_envelope_additions/.github/workflows/build-test.yml#L9) on CI. While we're at it, we should probably add 3.10 and 3.11 to that list instead.

examples/townhouse_walls_v3.xml
```diff
@@ -134,6 +137,21 @@
...
+                        </WallType>
+                        <Orientation>west</Orientation>
+                        <Insulation>
+                            <SystemIdentifier id="wall4ins"/>
+                            <Layer>
+                                <InstallationType>cavity</InstallationType>
+                                <NominalRValue>0</NominalRValue>
+                            </Layer>
+                        </Insulation>
+                    </Wall>
```
bpark1327, line 154: @nmerket Do you expect to always see at least 4 walls in HPXML? Or do we need to make assumptions when there are fewer than 4 walls?

nmerket: No, there can be any number of walls in HPXML. See [these docs](https://hescore-hpxml.readthedocs.io/en/v7.0.2/translation/zone_wall.html#wall-orientation) for how we try to flexibly support many different wall configurations.

bpark1327: See https://github.com/NREL/OpenStudio-HEScore/pull/385#issuecomment-1453923621

examples/hescore_min_v3.json
nmerket, line 1: We should remove this file. This should be equivalent to hescore_min.json. It's being tested against that file. Likewise we should remove all the house*_v3.json files and the house1-v2*.json files as they should be equivalent to the other ones.

We should update the create_all_example_json.py file to only create the ones we expect for the tests.

hescorehpxml/base.py
```diff
@@ -194,82 +195,100 @@ def wall_round_to_nearest(*args):
...
         # R-value and construction code
         if assembly_eff_rvalue is not None:
-            closest_wall_code, closest_code_rvalue = min(
-                [(doe2code, code_rvalue)
-                 for doe2code, code_rvalue in self.wall_assembly_eff_rvalues.items()
-                 if doe2code[2:4] == wallconstype and doe2code[6:8] == sidingtype],
-                key=lambda x: abs(x[1] - assembly_eff_rvalue)
-            )
-            return closest_wall_code, assembly_eff_rvalue
+            if is_exterior_wall:
```
nmerket, line 247: What happens if the wall _isn't_ an exterior wall? It looks like it could still try to make a construction code and it an error because `sidingtype` doesn't get set.

hescorehpxml/base.py
```diff
@@ -1795,11 +1804,25 @@ def get_building_zone_wall(self, b, bldg_about):
         hpxmlwalls = dict([(side, []) for side in list(sidemap.values())])
         hpxmlwalls['noside'] = []
         for wall in self.get_hescore_walls(b):
-            assembly_code, assembly_eff_rvalue = self.get_wall_assembly_code_and_rvalue(wall)
+            wall_id = xpath(wall, 'h:SystemIdentifier/@id', raise_err=True)
+            try:
+                # Require walls facing each direction with the ExteriorAdjacentTo
+                wall_adjacent_to = xpath(wall, 'h:ExteriorAdjacentTo/text()', raise_err=True)
+            except ElementNotFoundError:
+                raise TranslationError('%s has no orientation information.' % (wall_id))
```
nmerket, line 1812: In a lot of cases you don't need to catch the `ElementNotFoundError` as it will return the xpath of the element it was looking for and didn't find in the error message. It's subclassed from `TranslationError` so it will be caught at the higher level and returned without a gross Traceback.

hescorehpxml/base.py
```diff
@@ -1920,54 +1956,64 @@ def get_building_zone_wall(self, b, bldg_about):
...
+        # if bldg_about['shape'] == 'town_house':
+        #     if all_walls_same:
+        #         # Check to make sure the windows aren't on shared walls.
+        #         window_on_shared_wall_fail = windows_are_on_shared_walls()
+        #         if window_on_shared_wall_fail:
+        #             # Change which walls are shared and check again.
+        #             if bldg_about['town_house_walls'] == 'back_right_front':
+        #                 bldg_about['town_house_walls'] = 'back_front_left'
+        #                 window_on_shared_wall_fail = windows_are_on_shared_walls()
+        #         if window_on_shared_wall_fail:
```
nmerket, line 1988: Can we delete this commented out code?

hescorehpxml/base.py
```diff
@@ -1985,7 +2031,7 @@ def windows_are_on_shared_walls():
             heswall['zone_window'] = zone_window
 
             # If there are no windows on that side of the house
-            if len(windows) == 0:
+            if len(windows) == 0:  # FIXME: Do we need this?
```
nmerket, line 2034: You're right. It looks like this is no longer necessary. In OpenStudio-HEScore we [check if the side has windows](https://github.com/NREL/OpenStudio-HEScore/blob/d7400f92f128659d9bd7c2704c0e9e0c774a974c/rulesets/HEScoreRuleset/resources/HESruleset.rb#L384) and skip them if the `zone_window` key isn't there. I think the better thing to do would be to just omit the `zone_window` key if there are no windows.

hescorehpxml/hpxml2.py
```diff
@@ -184,3 +184,16 @@ def get_duct_location(self, hpxml_duct_location, v3_bldg):
...
+
+    enclosure_adjacent_to_map = {'ambient': 'outside',
+                                 'garage': None,
+                                 'attic': None,
+                                 'crawlspace': None,
+                                 'ground': None,
+                                 'living space': None,
+                                 'unconditioned basement': None,
+                                 'other housing unit': 'other_unit',
+                                 'other': None}  # FIXME: Check it
```
nmerket, line 199: What happens with `None` here?

hescorehpxml/lookups/make_lu_eff_rvalue.py
```diff
@@ -57,6 +57,29 @@ def main():
             assembly_u_value = 1 / assembly_r_value
             csv_writer.writerow([assembly_code, f"{assembly_u_value:.3f}", f"{assembly_r_value:.1f}"])
 
+    int_wall_asssembly_codes = ["iwwf00", "iwwf03", "iwwf07", "iwwf11", "iwwf13", "iwwf15", "iwwf19", "iwwf21"]
```
nmerket, line 60: For knee walls I'm pulling these from the json schema. I think it'd be good to do that here too so we don't have to maintain this list in two places.

hescorehpxml/schemas/hescore_json.schema.json
```diff
@@ -2642,7 +2274,29 @@
...
+                                "zone_wall": {
+                                    "allOf": [
+                                        {"$ref": "#/definitions/def_front_wall_constraint"},
+                                        {"$ref": "#/definitions/def_back_wall_constraint"},
+                                        {"$ref": "#/definitions/def_right_wall_constraint"},
+                                        {"$ref": "#/definitions/def_left_wall_constraint"},
+                                        {"$ref": "#/definitions/def_front_window_constraint"},
+                                        {"$ref": "#/definitions/def_back_window_constraint"},
+                                        {"$ref": "#/definitions/def_right_window_constraint"},
+                                        {"$ref": "#/definitions/def_left_window_constraint"}
```
nmerket, line 2291: Do these still apply?

tests/test_translation.py
nmerket, line 1: I see removing a lot of tests that aren't relevant anymore, but I'd like to see some new tests, especially of building types that we are adding: mobile homes and apartments.

nmerket (changes requested):
See my comments and questions below. Looks mostly good, I'd like to see some more testing of the capabilities we've added such test files for mobile homes and apartments and a couple permutations of the input possibilities there.

hescorehpxml/hpxml2.py
bpark1327: I can change it to raise TranslationError when it returns None. What do you think?

nmerket: I'm a little lost in the context, but yeah, if there's no other reasonable assumption we should do here. Whatever you decide make sure this ends up in the translation documentation.

docs/source/translation/zone_wall.rst
```diff
@@ -237,3 +237,33 @@ Then the nearest discrete R-value to the weighted average R-value from the looku
...
+
+.. table:: HPXML ExteriorAdjacentTo to HEScore adjacent_to mapping (HPXML v3)
+
+   ==========================  ====================
+   HPXML                       HEScore
+   ==========================  ====================
+   outside                     outside
+   other housing unit          other_unit
+   other housing unit above    other_unit
+   other housing unit below    other_unit
```
shorowit, line 267: Doesn't seem like these two enumerations should be used for a _wall_; they only make sense for a floor or ceiling.

hescorehpxml/schemas/hescore_json.schema.json
bpark1327: @nmerket I gave another thought to this. I think we still need to check all of these. But these constraints are overly complicated. I added these to zone_wall level https://github.com/NREL/hescore-hpxml/blob/29df54370a078b6425d2140087f6838e1fae2de1/hescorehpxml/schemas/hescore_json.schema.json#L1058-L1114
It's not going to return user-friendly error messages as before but I think it is sufficient.

PR #219, opened 2023-03-16
Support hpxml 3.1

yzhou601:
Fixes #

## Pull Request Description

description here

## Checklist

- [ ] Update translation docs

### DIFF
--- a/hescorehpxml/base.py
+++ b/hescorehpxml/base.py
@@ -181,7 +181,8 @@ class HPXMLtoHEScoreTranslatorBase(object):
                      'fiber cement siding': 'wo',
                      'composite shingle siding': 'wo',
                      'masonite siding': 'wo',
-                     'other': None}
+                     'other': None,
+                     'none': None}
 
         def wall_round_to_nearest(*args):
             try:
@@ -326,9 +327,6 @@ class HPXMLtoHEScoreTranslatorBase(object):
         glass_type = xpath(window, 'h:GlassType/text()')
         is_hescore_dp = self.check_is_doublepane(window, glass_layers)
         is_storm_lowe = False
-        window_frame = None
-        window_layer = None
-        window_glass_type = None
 
         if is_hescore_dp:
             # double pane needs more information being analyzed to determine glass type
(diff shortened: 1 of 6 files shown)


yzhou601:
@nmerket I went through the v3.1 release notes and made updates to the translator. I think this is ready for your review.

hescorehpxml/base.py
```diff
@@ -465,17 +465,22 @@ def get_heating_system_type(self, htgsys):
...
+            if len(eff_units) != 0:
+                eff_els = []
+                for eff_unit in eff_units:
+                    eff_el = htgsys.xpath(
+                        '(h:AnnualHeatingEfficiency|h:AnnualHeatEfficiency)[h:Units=$effunits]',
+                        namespaces=ns,
+                        effunits=eff_unit
+                    )
+                    if len(eff_el) > 0:
+                        eff_els.append(eff_el[0])
```
nmerket, line 483: Say an HPXML file has a heat pump with an HSPF and HSPF2. This looks like it will gather the first of each and put them in the `eff_els` array. Then further down you're picking the first item in `eff_els` and using that. That means that the efficiency sent to HEScore depends on whether HSPF or HSPF2 is first in the file. It should be more deterministic. How about some logic like:

- If there's an HSPF in the list of efficiencies on the heat pump, use that.
  - If there's more than one HSPF, use the first one.
- If there's an HSPF2 in the list of efficiencies on the heat pump, convert and use that.
   - If there's more than one HSPF2, use the first one.

hescorehpxml/base.py
```diff
@@ -342,6 +340,8 @@ def get_window_code(self, window):
         elif glass_layers == 'triple-pane':
             window_layer = 'triple-pane'
             window_glass_type = 'insulating low-e argon'
+        else:
+            raise TranslationError('Unhandled glass layers: {}'.format(glass_layers))
```
nmerket, line 344: Is there a change in HPMXL v3.1 that made it necessary to catch this, or is this something that we should've been catching all along and weren't?

nmerket (changes requested):
It looks like the changes were pretty minor between 3.0 and 3.1, which is good. It looks like they amount to

- Adding HSPF2/CEER/SEER2
- Something with windows. Could you elaborate on what that was about or point me in the right direction?

This still needs updates to the translation documentation and some tests exercising the changes. See the PR checklist for all the steps.

